PR #3, opened 2012-12-19
Python 3 fixes (keeping Python 2 compatibility)

cdeil:


piti118:
Python 3 is now supported in master

PR #15, opened 2012-12-20
WIP: Check notebooks on travis-ci

cdeil:


piti118:
Found this https://github.com/bollwyvl/nosebook

HDembinski:
What is the status of this PR? Shall we merge it? Shall we close it? Seems there is some problem with Travis.

cdeil:
I'm closing this old PR now.

@HDembinski - If you have time to set it up, IMO adding a build that executes the tutorial notebooks would still be helpful.

In Gammapy we use https://github.com/paulgb/runipy driven by a separate script: https://github.com/gammapy/gammapy/blob/master/test_notebooks.py

There's also https://github.com/computationalmodelling/nbval if we want to run it through pytest.

Overall I have to say I'm not a huge fan of the current iminuit docs. Almost all examples are in the notebook, but they are very long and it's hard to find something in there, the examples are not cross-linked with the API documentation.
In Gammapy we have integrated the notebooks and the Sphinx docs
(see http://docs.gammapy.org/dev/tutorials.html#notebooks) via https://nbsphinx.readthedocs.io
This way you get full-text search and table of content and navigation for the content in notebooks.

I don't have time to rework the iminuit docs, but @HDembinski if you have some time, my suggestion would be to either move more content into the Sphinx docs, or to go for shorter notebooks, ideally integrated with the sphinx docs via nbsphinx. But that's a lot of work, and not really high priority, so I'm not opening an issue with these suggestions.

PR #143, opened 2015-07-05
fixed import of StringIO

omazapa:
just fixed a small bug found for new versions of python.


piti118:
Thanks. Cython error was fixed in #145.

Issue #123, opened 2014-08-16
Make Python 2 and 3 console formatting consistent


iRaffnix:
Dear all!

unfortunately iminuit is not Python3 compatible.
Is it planned to overcome this little problem in the future?

Cheers,
iRaffnix

mattngc:
Hi, has there been any movement on this issue? Python 3 support for iminuit would be fantastic.

piti118:
Okie. I'll investigate this for real this time.

piti118:
OK Done. Try it out. It "should" work. There might be a few kinks here and there please report those as well.

kbarbary:
@piti118 Thanks so much. This is the last package I use that has been preventing me from switching to using Py3 full time. I'll report any problems I find.

mattngc:
@piti118 Yes, thanks so much for this. I've being doing some preliminary testing today. So far I can't find any issues. But then I'm not an advanced user, just doing simple stuff so far. But no problems so far. Thanks again.

mattngc:
This is working fine for me so far in python 3. Something I did notice however is that the console output is misaligned for values with exponents. To see this effect, copy code below into a text editor with fixed column widths. This might be a bug due to python 3 string handling being different to what you had before in python 2.

---------------------------------------------------------------------------------------------------------------------
|      |             Name             |  Value   | Para Err |   Err-   |   Err+   |  Limit-  |  Limit+  |          |
---------------------------------------------------------------------------------------------------------------------
|    0 |        Dalpha-cold_intensity =  6.913E+17 |  5.77E+13 |          |          |  5E+16   |  2E+18   |          |
|    1 |      Dalpha-cold_temperature =  41.29   |  0.003103 |          |          |  0       |  100     |          |
|    2 |       Dalpha-cold_wavelength =  655.9   |  3.965E-06 |          |          |  654     |  659     |          |
|    3 |   Dalpha-luke-warm_intensity =  2.608E+17 |  7.741E+13 |          |          |  0       |  6.071E+18 

Cheers, matt

cdeil:
@mattngc – Yes, float formatting changed on Python 3. I think it would be best if we make all our output be the same on Py 2 / 3 by being explicit with the formatting everywhere. If someone here can help out with this and make a PR, please leave a comment. Otherwise I'll try to get it done tomorrow.

Changing the title of this PR now and leaving it open for this formatting issue now.

mattngc:
[...]

PR #147, opened 2015-07-25
Python23 compat

piti118:


piti118:
#123 At least it compiles and pass all the tests.

PR #157, opened 2015-08-01
Code formatting and cleanup of issues pointed out by PyCharm

cdeil:
This is mostly PEP8 whitespace changes, there should be no change in behaviour.

I quickly browsed the diff on Github and it looks OK to me, so I'll just merge if travis-ci passes.


iminuit/_libiminuit.pyx
```diff
@@ -24,12 +25,12 @@ __all__ = ['Minuit']
 # Our wrapper
 cdef extern from "PythonFCN.h":
     #int raise_py_err()#this is very important we need custom error handler
-    FunctionMinimum* call_mnapplication_wrapper(\
-        MnApplication app,unsigned int i, double tol) except +
+    FunctionMinimum*call_mnapplication_wrapper( \
```
cdeil: @piti118 I think I screwed up here ... the PyCharm code formatting removed a space which is probably important!?

Why didn't this make the tests fail?

Of course we should revert this, but at the same time, we should add tests to make sure this is covered ...

PR #152, opened 2015-08-01
Fix Minuit.np_matrix

cdeil:
The [iminuit.Minuit.np_matrix](http://iminuit.readthedocs.org/en/latest/api.html#iminuit.Minuit.np_matrix) member function is broken:
```
    def np_matrix(self, correlation=False):
        """Error or correlation matrix in numpy array format."""
        import numpy as np
        #TODO make a not so lazy one
        return np.array(matrix)
```

I'll fix it and add tests.


cdeil:
This was easy ...

Issue #166, opened 2015-08-10
non-optional dependence on IPython


kbarbary:
With the latest master build but not the latest release version from pip (v1.1.1), I find a non-optional dependence on IPython being installed. To reproduce, install iminuit into a fresh environment with minimal dependencies (I used `conda create -n testiminuit numpy cython`), install iminuit from master, and try the example:

```python
>>> from iminuit import Minuit
>>> def f(x, y, z):
...     return (x - 2) ** 2 + (y - 3) ** 2 + (z - 4) ** 2
... 
>>> m = Minuit(f)
Traceback (most recent call last):
  File "iminuit/_libiminuit.pyx", line 1304, in iminuit._libiminuit.Minuit._auto_frontend (iminuit/_libiminuit.cpp:19288)
NameError: name '__IPYTHON__' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iminuit/_libiminuit.pyx", line 296, in iminuit._libiminuit.Minuit.__init__ (iminuit/_libiminuit.cpp:3048)
  File "iminuit/_libiminuit.pyx", line 1308, in iminuit._libiminuit.Minuit._auto_frontend (iminuit/_libiminuit.cpp:19377)
  File "/home/kyle/.conda/envs/testiminuit/lib/python3.4/site-packages/iminuit-1.1.1-py3.4-linux-x86_64.egg/iminuit/frontends/__init__.py", line 4, in <module>
    from .html import *
  File "/home/kyle/.conda/envs/testiminuit/lib/python3.4/site-packages/iminuit-1.1.1-py3.4-linux-x86_64.egg/iminuit/frontends/html.py", line 5, in <module>
    from IPython.core.display import display_html
ImportError: No module named 'IPython'
```

On python 2.7, I get the traceback

```
>>> m = Minuit(f)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iminuit/_libiminuit.pyx", line 296, in iminuit._libiminuit.Minuit.__init__ (iminuit/_libiminuit.cpp:3048)
  File "iminuit/_libiminuit.pyx", line 1308, in iminuit._libiminuit.Minuit._auto_frontend (iminuit/_libiminuit.cpp:19377)
  File "build/bdist.linux-x86_64/egg/iminuit/frontends/__init__.py", line 4, in <module>
[...]

PR #169, opened 2015-09-26
small changes needed to compile on windows, python 3.5, plus typo fixes in README

energynumbers:
two explicit includes were needed in the Minuit source.

The encoding at the top of setup.py needs to be "utf-8" with a hyphen, not without, in order to run at all on windows python 3.5.

I removed the import of html.py in `frontends/__init__.py`, as per #166 

And I fixed a couple of typos in the readme, and added explicit mention of the dependency on cython.


piti118:
Thanks. I also invited you to be a maintainer of this repo so you can merge as you see fit.

PR #181, opened 2016-02-10
typo in license statement

fabianrost84:


cdeil:
@fabianrost84 - Thanks!

The same line was copy and pasted to `doc/index.rst`, I've copied your fix there, too, in b3fb666.

Issue #175, opened 2015-12-16
Analytical derivatives?


giacomov:
Is there a way to specify the analytical derivatives of the function to minimize?

For simple functions this speed up a lot the fitting, and it is a feature implemented in Minuit2.

cdeil:
This is the same feature request as #125.
As far as I know, this Minuit2 functionality hasn't been exposed in iminuit and no-one's working on it at the moment.

cdeil:
This was implemented by @chrisburr in #182.
Closing this issue.

Everyone: please try it out on your applications and give feedback.

Issue #125, opened 2014-11-03
Wrap FCNGradientBase


cdeil:
It looks like one important feature of `Minuit` that is not exposed by `iminuit` is the possibility for the user to compute gradients themselves (see http://root.cern.ch/root/html/ROOT__Minuit2__FCNGradientBase.html).

@piti118 Do you think it would be useful to add this? Any thoughts on how to expose this to the end user?

piti118:
OK noted. This is an important feature. I want to make sure python3 support works first and I'll do this one.

iamholger:
Dear developers, 

in our project (professor.hepforge.org) we switched from pyminuit to iminuit entirely and are very happy with it. The above mentioned feature would benefit our project as we can often construct the gradient of the function we want to minimise. Cpu time of the minimisation currently limits the application to ~ 10 dimensional parameter spaces so it would be cool to go beyond that. 
Do you have an estimate on whether that MINUIT functionality will be part of iminuit anytime soon?

Thanks,
Holger

cdeil:
This was implemented by @chrisburr in #182.
Closing this issue.

Everyone: please try it out on your applications and give feedback.

iamholger:
Thanks a lot for providing this functionality. The usage is nice and simple. I did a few consistency checks and found the minimisation results with numerical and analytical gradients to be consistent. 
I will do more testing soon. 

We have some output in our minimisation stage that looks like that now:
 ******************************************************************
* Using iminuit, please visit https://github.com/iminuit/iminuit *
*******************************************************************

Let me know if you want something else or more verbose.

Thanks a lot!

PR #182, opened 2016-02-10
Add support for supplying the gradient function to minuit

chrisburr:
This PR addresses #125 and #175 and adds the `grad_fcn` keyword  argument when initialising `Minuit`.

Example usage:
```python
def f(x, y, z):
    return 0.2 * (x - 2.) ** 2 + 0.1 * (y - 5.) ** 2 + 0.25 * (z - 7.) ** 2 + 10
def f_grad(x, y, z):
    dfdx = 0.4 * (x - 2.)
    dfdy = 0.2 * (y - 5.)
    dfdz = 0.5 * (z - 7.)
    return (dfdx, dfdy, dfdz)

m = Minuit(f, grad_fcn=f_grad)
m.migrad()
```

### DIFF
--- a/iminuit/Minuit2.pxi
+++ b/iminuit/Minuit2.pxi
@@ -14,6 +14,14 @@ cdef extern from "Minuit2/FCNBase.h":
         double call "operator()"(vector[double] x) except+
         double ErrorDef()
 
+cdef extern from "Minuit2/FCNGradientBase.h":
+    cdef cppclass FCNGradientBase:
+        FCNGradientBase(object fcn, double up_parm, vector[string] pname, bint thrownan)
+        double call "operator()"(vector[double] x) except +  #raise_py_err
+        double Up()
+        vector[double] Gradient(vector[double] x) except +  #raise_py_err
+        bint CheckGradient()
+
 cdef extern from "Minuit2/MnApplication.h":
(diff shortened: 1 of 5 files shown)


cdeil:
@chrisburr - I don't have time to try this right now, but I will in the coming days. Thank you!

@piti118 @iamholger @gitcheng @giacomov @aarchiba or whoever might be interested in analytical gradients for iminuit: do you have time to review API and / or implementation or try to apply this for an example you care about and give feedback within the next ~ week?

cdeil:
@chrisburr - If you have time, could you please add docs for this?

IMO a separate small page `doc/gradient.rst` that contains an example (like the one you showed above) where providing gradients helps would be best. I.e. run MIGRAD (and maybe one of the other steps like minos or contours where gradient is used) without and with gradient, and mention the number of function evaluations for each case.

If you think this is too extensive, adding the gradient example here after the example without gradient would also be good, so that users find out about this feature: https://github.com/iminuit/iminuit/blob/master/doc/index.rst#in-a-nutshell

cdeil:
Merging this now. Thanks @chrisburr!

I'll add an example to the docs in a separate PR.

piti118:
This should appear in tutorial notebook too.

Issue #189, opened 2016-11-02
Attribute error when calling `migrad`


xaratustrah:
Hi and thanks for the nice work. I am not sure if this is a bug, but I encounter an **Atribute error** when calling `migrad` using `*` parameters. I have put the error description [here on SO](http://stackoverflow.com/q/40360615/5177935). Any hints is appreciated.

cdeil:
@xaratustrah - Thanks for reporting this.

Please see http://stackoverflow.com/a/40389851/498873

I think for iminuit there's two action items here:

- [ ] Implement input validation in `Minuit.__init__`
- [ ] Add one or more documentation examples that use `forced_parameters` to avoid argument inspection.

I don't have time for this at the moment.
Pull requests welcome!

cdeil:
Just to be clear: the cause of the error is that you're passing a float `frontend=0.4` to `Minuit.__init__`, which gets stored in `self.frontend` here:
```
self.frontend = self._auto_frontend() if frontend is None else frontend
```
https://github.com/iminuit/iminuit/blob/master/iminuit/_libiminuit.pyx#L343

and then later this line
```
self.frontend.print_banner('MIGRAD')
```
https://github.com/iminuit/iminuit/blob/master/iminuit/_libiminuit.pyx#L439

causes the `AttributeError`:
```
$ python fit.py 
57.1645229329
fit.py:16: InitialParamWarning: errordef is not given. Default to 1.
  m = iminuit.Minuit(minimize_me, *p)
Traceback (most recent call last):
  File "fit.py", line 17, in <module>
    m.migrad() # fails!
  File "iminuit/_libiminuit.pyx", line 439, in iminuit._libiminuit.Minuit.migrad (iminuit/_libiminuit.cpp:5074)
AttributeError: 'float' object has no attribute 'print_banner'
```

So it's clear what to do: add a regression test for this and then input validation in `Minuit.__init__`.

xaratustrah:
Thank you for taking care of this so far. I am not sure if the following suggestion would be useful and maybe easier. Since parameter names are needed and actually quite good in **iminuit**, in a manner _similar_ to `scipy.optimize`:

```
from scipy.optimize import curve_fit
p = [1,2,3,4,5]
curve_fit(fit_function, x_data, y_data, p0 = p)
```

[...]

PR #190, opened 2016-11-04
Added a classmethod as an alternative constructor.

xaratustrah:
@cdeil - based on your suggestion and the [discussion on SO](http://stackoverflow.com/a/40389851), a simple implementation of a classmethod, which works well as far as I tested. I hope this PR can solve issue #189. Here some code for testing:
```
import iminuit
import numpy as np

x_data = np.array([0,1,2,3,4,5,6,7,8,9])
y_data = np.array([0,1,2,3,4,5,4,3,2,1])

def fit_function(x, *p):
    return p[0]*np.exp(-((x-p[1])**2/(2*p[2]**2)))

def minimize_me(*p):
    return sum((fit_function(x, *p) - y)**2 for x, y in zip(x_data, y_data))

p_vals = [4.5, 5, 0.4]
m = iminuit.Minuit.from_parameter_list(minimize_me, p_vals, pedantic=0, print_level=2)
result = m.migrad()
print(m.parameters)
```
With kind regards, @xaratustrah.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -38,6 +38,20 @@ In a nutshell
     m.migrad()
     print(m.values)  # {'x': 2,'y': 3,'z': 4}
     print(m.errors)  # {'x': 1,'y': 1,'z': 1}
+    print(m.parameters) # ['x', 'y', 'z']
+
+Or alternatively using a list of estimates:
+
+.. code-block:: python
+
+    from iminuit import Minuit
+    def f(*p):
+        return (p[0] - 2) ** 2 + (p[1] - 3) ** 2 + (p[2] - 4) ** 2
+    m = Minuit.from_parameter_list(f, [20,30,40])
+    m.migrad()
+    print(m.values)  # {'x': 2,'y': 3,'z': 4}
+    print(m.errors)  # {'x': 1,'y': 1,'z': 1}
+    print(m.parameters) # ['p0', 'p1', 'p2']
 
 If you are interested in fitting a curve or distribution, take a look at `probfit`_.
 
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -398,6 +398,16 @@ cdef class Minuit:
 
         self.merrors_struct = {}
 
+    @classmethod
+    def from_parameter_list(cls, fcn, params_list, **kwargs):
+        """
+        Alternative constructor. Gives default names to parameters.
+        """
+
+        p_names = ['p{}'.format(_) for _ in range(len(params_list))]
+        obj = cls(fcn=fcn, forced_parameters=p_names, **dict(zip(p_names, params_list)), **kwargs)
+        return obj
+
     def migrad(self, int ncall=10000, resume=True, int nsplit=1, precision=None):
         """Run migrad.
 
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -400,8 +400,25 @@ cdef class Minuit:
 
     @classmethod
     def from_parameter_list(cls, fcn, params_list, **kwargs):
-        """
-        Alternative constructor. Gives default names to parameters.
+        """Alternative constructor.
+        Gives default names to parameters, in case these are not specifically given. Internal
+        forced_parameters argument is set. This style of instantiation is useful, whenever
+        explicit naming is either not possible or not desired.
+
+        **Arguments:**
+
+            - **fcn**: the function to be optimized. Minuit automagically finds
+              parameters names. More information about how
+              Minuit detects function signature can be found in
+              :ref:`function-sig-label`
+
+            - **param_list**: A list of values for the parameters. A name will be automatically
+            assigned to them in the same order i.e. p0, p1, p2, ...
+
+        **Parameter Keyword Arguments:**
+
+            - These will be passed to the default constructor **Minuit.__init__()**
+
         """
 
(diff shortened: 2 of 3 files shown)


cdeil:
@xaratustrah - Thanks!

---

Concerning implementation / API:

I think what's proposed here is helpful, +1 to add it.

A larger change where we deprecate `forced_parameters` in `Minuit.__init__` and move all it's functionality (including the option to give a list of parameter names) in this constructor or a third constructor would be better. Should we discuss / consider this larger change at all / here?

--- 

@xaratustrah - Can you please add at least one doc example for this?

I find the example you post above has some complexity that's not really needed to test / document this.
Maybe just add a second example here, that executes the same or a similar example, but using this new constructor?
http://iminuit.readthedocs.io/en/latest/#in-a-nutshell

The docstring should describe each parameter and what is returned, and mention that `**kwargs` are forwarded to `Minuit.__init__`, i.e. be very explicit how it works.

---

There should be at least one test that runs the docs example you add.
The implementation is pretty simple, so I actually think one or a few simple tests are enough.

cdeil:
@piti118,  and all interested in iminuit - please comment on this PR, and the question of what you'd like the long-term solution for `forced_parameters` to be.

xaratustrah:
@cdeil - As suggested, I have updated the docstrings, added a simple doc example and also functions for **pytest**.

At some point `sphinx` started to complain about a missing blank line while generating the docs, that's why the TravisCI Check is showing failure. All other tests pass. I have been googling, but I have still no idea, what exactly the error is about:

```
Warning, treated as error:

docstring of iminuit.Minuit.is_clean_state:0: WARNING: Bullet list ends without a blank line; unexpected unindent.
```

Funny enough, the error seems to related to a [section of the code](https://github.com/iminuit/iminuit/blob/master/iminuit/_libiminuit.pyx#L852) which I actually didn't change at all.

Any insights are welcome. With kind regards, @xaratustrah.

HDembinski:
Hi, can we revive this PR? What is blocking approval? It still sounds useful. I found myself in a similar situation before, and this solution is much more elegant than the one based on `forced_parameters`.

HDembinski:
I have been doing some reading into this PR and discussed the issue with @cdeil offline as well.

I agree with everything that was said here so far and I also think that it would be very handy to use Minuit in a way that is more compatible with (https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize)[scipy.optimize.minimize], where you don't specify a name for each parameter, because it works better when you want to play around with the number of fitted parameters.

I think we have two options.

**Fix this PR and merge it**
The solution and the interface of the new classmethod are straight forward, the broken build is due to some seemingly trivial bug in the RST. This should be easy to fix. We could even deprecate the use of `forced_parameters` with a warning message.

**Alternative: add a minimize function to iminuit module**
We could offer a `minimize` function in the `iminuit` module, which mimics the interface of `scipy.optimize.minimize` and provides a compatible output. The advantage would be that you can very easily switch between scipy and iminuit in your script, and try both optimizers. Naturally, this call also accepts a list of parameters, but also requires a different, more numpy-compatible way of calling the objective function.
```
# iminuit style objective function with variable number of arguments
def obj1(*parameters):
   # parameters is a tuple of parameter values
    ...

# scipy style objective function
def obj2(parameters, *args):
   # parameters is an array of parameter values
   # args is a tuple with auxiliary fixed parameters to control the function
   ...
```
Now that I write this down, I don't like it as much anymore, at least not as an alternative. It is a related feature, but perhaps goes farther than what @xaratustrah wants. It could be a separate feature.

HDembinski:
Ok, I gave it more thought and I think the interface needs to be enhanced. It does not support initial step sizes and parameter limits.

According to the current implementation, you can only do
```
     m = Minuit.from_parameter_list(f, [20,30,40]) 
```
But what if you want to specify the initial step size, what is called `error_<parameter name>` in `Minuit.__init__`. We need an elegant solution for that as well that does not fall back to parameter names. Similarly for limits.

If I want to set initial step sizes, I would currently have to do
```
    m = Minuit.from_parameter_list(f, [20, 30, 40], error_p0=5, error_p1=5, error_p2=5)
```
That is long and inelegant. The parameter names p0, p1, and p2 are an implementation detail of `from_parameter_list`, which should not leak back into the interface like this. The user should not need to know what names Minuit internally chooses.

I suggest something like this
```
   m = Minuit.from_parameter_list(f, [20, 30, 40], error=[5, 5, 5], limit=[(0, 100), (0, 100), (0, 100)])
```
which would be in concordance with the naming scheme of the `Minuit.__init__`.

cdeil:
Some people are thinking about optimiser interface in scipy:
https://mail.python.org/pipermail/scipy-dev/2018-March/022565.html
I just wanted to leave this as a note, I don't have time to read that this week.

HDembinski:
Adapting iminuit to a general optimizer interface would be nice, but it probably would require modifications to the original Minuit code. The reason is that Minuit has hardwired stopping criteria, the EDM criterion and total number of function calls. Neither of these corresponds to the general stopping criteria used in scipy or nlopt, which mainly use absolute or relative convergence of the objective function value.

HDembinski:
This PR can be closed, it is superseded by my new PR

cdeil:
I'd suggest to keep this PR open until #229 or some form of this feature is merged.

@xaratustrah - If you have time, please look at or try out #229 .

HDembinski:
I am closing this, as it is superseded by #229

PR #202, opened 2018-02-20
Fixes

HDembinski:
Removed trailing whitespace.
Fixed all fixable compiler warnings (only the ones in the auto-generated Cython file remain).
Fixed the failing draw_contour() test.

### DIFF
--- a/Minuit/inc/Math/Error.h
+++ b/Minuit/inc/Math/Error.h
@@ -21,21 +21,21 @@
 
 
 
-/** 
-   Pre-processor macro to report messages 
-   which can be configured to use ROOT error or 
+/**
+   Pre-processor macro to report messages
+   which can be configured to use ROOT error or
    simply an iostream in case of stan-alone builds
 */
 
 #ifndef USE_ROOT_ERROR
 
-// use iostream instead of ROOT 
+// use iostream instead of ROOT
 
 #include <iostream>
 
 #ifndef ROOT_MATH_LOG
 #define ROOT_MATH_OS std::cerr
-#else 
+#else
 #define ROOT_MATH_LOG
 #endif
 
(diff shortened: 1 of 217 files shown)


cdeil:
@HDembinski - Thanks!

Is it possible to get these whitespace / compiler warning fixes into the upstream Minuit package?
Or maybe they have already been applied there?

From what I understand, iminuit bundles `Minuit`, at the moment as a simple copy, without patches. I think before manually patching it on our side, updating to the latest upstream version (without any patches applied) would be better (especially since the fixes are only cosmetic, i.e. whitespace and compiler warnings, according to your commit message). That said, I never looked (or at least don't remember) at where the files in `Minuit` exactly come from. Probably `Minuit` is maintained since the last years in the `ROOT` repo?

Can you split out 3214a5c306243045f67bc851a6f20193ff32707b in a separate PR please? If this fixes a failing test we definitely want that in immediately.

HDembinski:
I do not see why the version of Minuit shipped with iminuit must be a hard copy of the original code. Especially if it is not clear which version of Minuit the "original code" is.

I agree with you that removing trailing white space is cosmetic, but fixing compiler warnings is not. The warnings were about narrowing conversions. Technically these are bugs that could produce buffer overflows and segfaults, even if they don't affect the logic of the code if it is used in the usual way. The important part is "in the usual way".

But even if you are not swayed by this argument, consider my second one, which is very practical. It is *very difficult* to develop code that throws so many warnings, because you miss the important ones that you introduce yourself during development.

I will split the patch as you suggested, but I ask you to consider accepting my other fixes as well.

HDembinski:
I close the request after the discussion we had offline. :)

alexpearce:
What action was agreed offline? Will the Minuit patches be submitted upstream?

HDembinski:
Hi @alexpearce, @cdeil convinced me that it makes more sense to patch Minuit upstream. Perhaps we are lucky and the upstream Minuit has fixes for these warnings already. I will check this. If upstream Minuit has fixes, then we should rather consider to merge the upstream Minuit package.

alexpearce:
Absolutely. Sounds like a reasonable plan.

cdeil:
I'm not sure, but I think maybe this is the place where `minuit2` is maintained, i.e. where we should copy from: https://github.com/root-project/root/commits/master/math/minuit2

PR #203, opened 2018-02-21
fixed failing test_drawcontour

HDembinski:
The v array used to label the contours. It was not in sync with the bound argument of the call to self.contour(...) which generates the contours. This is now fixed.


cdeil:
Thanks!

HDembinski:
@cdeil Cool, thank you! I am honored and will use this permission responsibly.

alexpearce:
No problem at all @cdeil. Thanks a lot for the contributions @HDembinski!

HDembinski:
@alexpearce No problem that one was easy :), now I am looking into updating Minuit code from ROOT, this is hard.

PR #204, opened 2018-02-22
Build improvements

HDembinski:
These patches make development more fluid and convenient. These are all backward-compatible improvements with no changes to users. All warnings should be gone now, if you use gcc or clang.

`make` now runs `make build`, not `make help`, which is most commonly what you want when you type `make`. I set the dependencies of the `build` target properly, so that the module is correctly rebuild when the sources change. This was not the case before.

I found a recipe on the internet to make setuptools use compiler-specific flags. This allows me to add flags for gcc and clang to ignore warnings in code generated by Cython and in the externally developed Minuit code. As it is, we cannot fix these warnings, they need to be fixed in the Cython code generator and the original Minuit source code. This is the best workaround for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.pyxbldc
 *.root
 .DS_Store
 *.so
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,16 @@
 PROJECT = iminuit
 CYTHON ?= cython
 
+default_target: build
+
 help:
 	@echo ''
 	@echo ' iminuit available make targets:'
 	@echo ''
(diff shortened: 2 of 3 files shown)


cdeil:
Added  `-Wno-sign-compare` so that I also get a clean build in e6b2600 .

This seems OK to me, merging now.

Personally, I would have preferred to not touch the build and add this complex code in setup.py, but instead to recommend this to devs that want a clean build in the dev docs:
```
export CFLAGS="-Wno-sign-compare -Wno-shorten-64-to-32 -Wblabla ..."
```
So if this should cause issues for anyone, I'd suggest that we revert to that.

Issue #193, opened 2016-11-21
using Minuit2 instead of SEAL minuit


giacomov:
Similar to #124 

I am working on a project where the user can choose between Minuit2 from the ROOT package and iminuit for minimization problems.

In our experience, the former often outperform the latter in finding better solutions to minimization problems. The reason for the difference is that the algorithm in Minuit2 has changed slightly since SEAL minuit, and new fixes have been added to the code:

https://root.cern.ch/gitweb/?p=root.git;a=history;f=math/minuit2/src;h=60885f144b2a0682bd358c23e3f8ebd766847fea;hb=HEAD

Since Minuit2 is released also as a standalone package and is free open source software (http://project-mathlibs.web.cern.ch/project-mathlibs/sw/Minuit2/html/index.html), why sticking with SEAL minuit which is not maintained anymore?

cdeil:
@giacomov - Thanks for bringing this up.

The top-level docs page indeed says that we're using SEAL Minuit:
http://iminuit.readthedocs.io/en/latest/#what-is-iminuit

But my understanding is that @piti118 in #145 updated to Minuit2 version 2.5 in July 2015, in response to #124.

Before we change the bundled version or docs page, I have some questions:

- So Minuit2 is now being maintained in the ROOT repo and [this](https://root.cern.ch/gitweb/?p=root.git;a=history;f=math/minuit2/src;h=60885f144b2a0682bd358c23e3f8ebd766847fea;hb=HEAD) is the changelog?
- What is the last stable version of standalone Minuit2 that was released?
- Are they still making stable versions or should we take versions of standalone Minuit2 or should we take versions from stable ROOT releases?

Does anyone know, or should we ask in the ROOT forum?

@piti118 - What is the procedure to update the bundled `Minuit2` for `iminuit`?
Is it just a `cp` command, or are there modifications we have to make? Can we document or script those, so that updating the bundled version becomes a simple and safe process that anyone can do?

giacomov:
Hi,
thanks!

I opened a question in the ROOT forum:

https://root.cern.ch/phpBB3/viewtopic.php?f=15&t=22784

Let's see what they say. Hopefully they will answer to most of your questions.

cdeil:
@giacomov - Thanks!

Please leave another comment here when you get some info.
(I can't remember my CERN passwd any more, so I can't easily watch the post.)

henryiii:
[...]

PR #205, opened 2018-02-22
New minuit2 code from ROOT 6.12.06, addressing issue #193

HDembinski:
This is addressing issue #193.

I pulled in the lastest Minuit2 code from ROOT. I had do some minor edits to remove ROOT things that had crept in. Note that the ROOT code is using C++11 features now, which means that we need to add the `-std=c++11` compiler flag in `setup.py`. In my other PR, `build_improvements`, I demonstrated a way to set different flags for different compilers, which we need to keep support for `msvc` going. We can use that trick here as well to set the compiler flags for C++11 support, which differ between `msvc` and the unix compilers.


cdeil:
We discussed offline. Plan is to first add a Windows CI build on Appveyor, and then try this update again, with update of upstream files and patches / changes separated into a few commits that make it simpler to re-do another update in the future by following the steps from this update.

Issue #165, opened 2015-08-06
'print_level=2' doesn't work in python 3


mattngc:
Levels 0 and 1 both work as expected, but not at level 2.

cdeil:
@mattngc – Print level for which method doesn't work as expected?

If you could produce a minimal code example that shows the issue, that would help a lot to pin this down and fix it quickly (and add a regression test to make sure it stays fixed). You're saying this gives the output you expect on Python 2 but not Python 3?

piti118:
This is because when we move from Minuit1.xx to Minuit2.xx I probably forgot to add some print function.

Print level 2 comes from Modified version of Minuit.

HDembinski:
@mattngc Hi, if you are still reading this, could you try the latest master version to whether your problem was solved?

mattngc:
Hi @HDembinski , I haven't used this library for a while. So not easy for me to test straight away. But thanks for looking at it.

cdeil:
I'm closing this issue now.

If anyone still finds an issue in the future related to this, please re-open or file a new one!

PR #206, opened 2018-02-27
New minuit v2

HDembinski:
Imports an up-to-date version of Minuit2 from ROOT-6.12.06 (this is the second attempt at doing this that's why it is called v2).

The import is done semi-automatically by a Python script, which can be found in the `Minuit` sub-directory to repeat the process. It contains detailed instructions of how the import was handled. Three files in Minuit2 need to be patched for iminuit to work, two of these patches are fixes that I will try to commit to upstream ROOT.

Minuit2 uses C++11 now, so unix compilers now need an additional flag `-std=c++11` for the code to compile. Thanks to the build system improvements done recently, it is easy to add this flag only for unix compilers.


cdeil:
@HDembinski - Thanks!

I would prefer to finish #207 first and have appveyor build and tests running here. Just as a practical matter, it seems that Windows build worked so far and we want to keep it working and I have no Windows box / no idea if these C++11 changes will work.

For reference: the upstream patches are proposed in https://github.com/root-project/root/pull/1677

I didn't review your update script and changes here yet.

Minuit/import_minuit2.py
```diff
@@ -0,0 +1,112 @@
...
+                    fp = src_files[key]
+                    imported[key] = fp
+                    if fp.inc is not None:
+                        scan(imported, fp.inc)
+                    if fp.src is not None:
+                        scan(imported, fp.src)
+
+    for k,v in imported.items():
+        if v == "NOT-FOUND":
+            print "Missing:",k
```
cdeil: Can you try the script on Python 3 once, please?
This print statement will fail.

I didn't read the script yet or try it - do you clear out the Minuit2 folder in our repo before copying new files?
I.e. is there a possibility that old, unused files are left lying around?
(this usually doesn't hurt, but in principle it could cause problems, depending on how the build works and picks up files)

HDembinski:
Sure, we want to keep the windows build working, that's why I set up the Appveyor CI. It makes sense to finish that first and let it test this branch. I guess I have to rebase feature branch once the the Appveyor PR is merged, so that it gets the `.appveyor.yml`. How do I do that?

piti118:
Awesome. Thanks a lot.

henryiii:
I've submitted a patch to allow Minuit2 to be built standalone again in root-project/root#1680 (and Windows builds in root-project/root#1681). Until then, https://github.com/GooFit/Minuit2 might be useful. I also have a proof-of-concept PyBind11 minuit2 wrapper in a branch on that repository, it includes a `setup.py` based build system for Minuit2 that might be useful.

HDembinski:
Hi @henryiii, your project is definitely aligned with our needs, thanks for tuning in. :)

HDembinski:
@henryiii I looked at your pybind11 wrapper, it is really nice.

HDembinski:
@cdeil That makes sense. Just for posterity, using pybind11 as an alternative to Cython has several advantages. I am a big fan of pybind11, it is an excellent library already and shows active progress with many cool features added.
- pybind11 is a header-only C++ library, it can be internalized into the project (as recommended by them), so it does not add an external dependency; in turn we could drop Cython as a dependency
- pybind11 is better than Cython for wrapping C++ libraries, it supports all C++ features, including all new C++11 features, while Cython does not, for example, Cython requires C++ classes that are wrapped in Python to have a default ctor if they are allocated on the stack

henryiii:
PyBind11 also doesn't require any learning a new language. I switched from Cython to PyBind11 for wrapping some time about and am continually amazed at how well it works. I converted the PyLandau from Cython to Pybind11 and it went from many lines to about 5 lines, and had better vectorization support. :) I've also hacked a bit on the internals of both, and PyBind is far clearer and simpler.

Also, CMake is not a dependency; I wrote a setup.py only build for Minuit2.

My PyBind11 bindings have a very specific purpose; I need GooFit's PyBind11 bindings to return Minuit2 classes, and then the user needs to be able to interact with them. PyBind allows that. I haven't verified it cross-module yet, though I think it is supposed to work. Passing C pointers, like iMinuit supports, should be possible too.

I'd like to talk with you @cdeil at some point about the design of iMinuit; I might be able to get funding for a student to work on filling out the PyBind11 classes at some point. I've been busy with other things, but it's been on my radar for a while...

HDembinski:
According to my manual tests, this patch also fixes issue #165

HDembinski:
@henryiii @cdeil Perhaps we should really organize a Skype call. I already dived into the iminuit internals deep enough to see that a switch from Cython to pybind11 would only require a bored weekend for me.

cdeil:
Are there any opinions if dropping Python 2.7 support on Windows only is an option for us?

I think the vast majority of scientists on Windows use Anaconda, and probably very few still need Python 2?

We could try to get some download stats info from
https://anaconda.org/conda-forge/iminuit/files
or PyPI.

HDembinski:
@cdeil I need to better understand how the build environment on Appveyor works, but it seems possible to me that we could fix this problem in several ways.
1) By separately compiling the Minuit code and the Cython-generated interface code. This is something we should do anyway, because the Minuit code rarely changes compared to the Cython-generated code. Always recompiling Minuit is not necessary. The Minuit code is C++11 and needs a C++11 compiler, while `setup.py` automatically requests Visual C++ 9.0 which is ancient but sufficient to compile the Cython-code only. This approach would work under the assumption that the libraries are binary compatible, which may not be the case.
2) By patching Minuit2 back to C++03. Probably the easiest route, since it is C++03 code anyway, the new features only crept in.

cdeil:
@HDembinski - I don't think mixing compilers on Windows (your option 1) is a good idea to pursue.
See  http://stevedower.id.au/blog/building-for-python-3-5/
I know close to nothing about Windows, and probably for you as well, so we don't want to spend our time there.

So I'd say either try to get it to compile on the normal, standard C++ Windows compiler for Python 2.7, or we just drop Python 2.7 support on Windows, and annoy a very small fraction of our users (1% 5% ?)

henryiii:
C++11 + Windows Python 2.7 is possible, in fact, it's required for PyBind11. See: http://pybind11.readthedocs.io/en/stable/faq.html#working-with-ancient-visual-studio-2009-builds-on-windows

HDembinski:
@henryiii Good point, I had the same idea and looked into their appveyor script. I think it works for them because they build with cmake and not setuptools. Anyway, I removed the C++11 code and it works now.

HDembinski:
@cdeil I think we are ready to merge now. :)

henryiii:
Especially since it's only one file, can you make a PR to upstream Minuit2?

cdeil:
@HDembinski - Thank you!

HDembinski:
@henryiii It turns out that I introduced this C++11 code myself, so there is no need to patch it in ROOT. I know, however, that there is more C++11 code in other parts of upstream Minuit2 which we do not include anymore, because we only copy files that are actually used by us. Fortunately, this means that we avoid this other C++11 code.

PR #207, opened 2018-02-27
Appveyor support

HDembinski:
Build iminuit and run the tests on Windows for Python 2.7, 3.4, and 3.6. The tests are run without installing Cython and IPython. A bug in test_describe.py was fixed, which simply assumed that Cython is always present.


HDembinski:
@cdeil Regarding `iminuit/_libiminuit.cpp`. Either we ship it and don't require Cython as a dependency, or we don't ship it and strictly require Cython as a dependency. This needs a decision. What you suggest is in contradiction with `setup.py`, where you can find the following comment:
```
# We follow the recommendation how to distribute Cython modules:
# http://docs.cython.org/src/reference/compilation.html#distributing-cython-modu
les
```
When you follow the link you will find the official Cython recommendation:

"It is strongly recommended that you distribute the generated .c files as well as your Cython sources, so that users can install your module without needing to have Cython available."

The code in setup.py was written so that you don't need Cython to build the project. I followed that lead. I also makes more sense to not require Cython. I like about iminuit that it is dependency-free in principle. To compile it, only a standard Python installation and a C++ compiler is needed. You only need Cython when you change any of the `.pyx` and `.pxi` files.

HDembinski:
@cdeil Regarding Appveyor status not showing: I don't know. This is only the second time I set up Appveyor and the first time it actually compiles and tests the project successfully. I noticed that I have to trigger the builds manually, they are not triggered automatically when I push to my iminuit fork.

Help is appreciated :)

cdeil:
What all other project that use Cython that I know of do it the following:

- The generated C is not stored in the repository. If someone wants to build the development version (usually only developers) they need to have Cython installed.
- When making a release, then `python setup.py sdist` generates the C file and it is included in the source tarball that is released on PyPI. I.e. users that `pip install iminuit` from source don't need Cython.



I'm +1 to go to this scheme also for `iminuit` now. Generally having auto-generated files in a git repo (especially such huge ones) is a bit painful, i.e. have them show up as huge diffs in pull requests (in this case a pull request on adding Appveyor, not initially with the goal to update the Cython-generated C output)

I don't know why it was set up in a different way initially. @piti118 - if you're still seeing these notifications: can you please comment? Are you OK to change to the more common scheme, i.e. not have the auto-generated C files in the repo?

cdeil:
I'm not seeing something at https://ci.appveyor.com/project/iminuit/iminuit

Did you set up the Appveyor project? How?
(I'm in a call now, we can do it together in 1 hour if you have time)

HDembinski:
Check out https://ci.appveyor.com/project/hdembinski/iminuit

HDembinski:
@cdeil Sorry for causing confusion here, the file `iminuit/_libiminuit.cpp` was not part of the master branch. I added it only in this branch, and I did that because of the comments in `setup.py`. What you suggest makes sense, so I will amend this branch by removing `_libiminuit.cpp` and adding Cython as a dependency in the appveyor build.

HDembinski:
@cdeil I setup Appveyor for iminuit, but PR commits do not trigger the build. Could you please check whether the iminuit account on github is connected properly to Appveyor? It is necessary to authorize Appveyor as a CI system, this can be done in the Settings in github. Also in the settings, please check whether this webhook for Appveyor appears
https://ci.appveyor.com/api/github/webhook?id=nsrc6d4viths3942

PR #208, opened 2018-02-27
dependency fix, improves performance

HDembinski:
Forgot to put the correct paths. It works without, but `make` runs some superfluous commands if the dependencies are set incorrectly.


cdeil:
Thanks!

PR #216, opened 2018-03-03
improvement of console frontend

HDembinski:
- removed what looks like accidental debug prints
- added a header explaining that the matrix shown is a correlation matrix and not the covariance matrix


cdeil:
@HDembinski - Thanks!

From the code diff, it's not obvious that this is better. In the future, please copy & paste the old and new output for one example in the issue description (or even an automated test) to make review simpler and to have a record what changed on Github.

But I presume that you've done that locally and that this change is good. Merging now.

HDembinski:
Ok, that makes sense. Sorry!

Issue #179, opened 2016-02-08
How to cite iminuit?


iamholger:
Hi,

we are happily using iminuit in our project  (professor.hepforge.org). We would like to cite iminuit properly but so far don't know how.

Thanks,
Holger

cdeil:
@piti118 @Dapid @energynumbers @fabianrost84 @lgeiger @omazapa You've contributed to iminuit in the past (well, @piti118 has written almost all of it). Do you have any opinion how iminuit should be cited?

I think the options are:
A) Ask people to add a link to https://github.com/iminuit/iminuit if possible
B) Put the 1.2 or a future release on https://zenodo.org/ to get a DOI
C) Someone volunteers to take the lead on writing a paper about iminuit in the near future.

I think C could be great (nice examples, benchmarks, technical details about C / Python / Cython), but I presume there's no-one that wants to put in that work in the coming months, right?

So do you think putting it on Zenodo is useful?

fabianrost84:
C would be great but I did almost nothing and understand nothing of the code really so I do not see myself there. I recently [published](http://elifesciences.org/lookup/doi/10.7554/eLife.10230.001) and cited iminuit using A. I don't know Zenodo and don't think a DOI is absolutely necessary. A is also [what scipy does](https://scipy.org/citing.html).

iamholger:
I'd suggest Option B as a short term solution, it's really quick to do and since zenodo is hosted at CERN there will be long term support.

piti118:
Would be nice if someone set up Zenodo.

fabianrost84:
Now that I started to use Zenodo myself I changed my mind. Zenodo would be great.

Dapid:
Another alternative for the paper: http://joss.theoj.org/

cdeil:
+1 for http://joss.theoj.org/ and to do a short writeup of what MINUIT iminuit is.

I remember that when I learned about MINUIT and iminuit I was super confused about:
- What algorithm MINUIT implements (http://iminuit.readthedocs.io/en/latest/#technical-stuff has good references)
- What version of MINUIT is the current one and used by iminuit
- How iminuit is implemented (the `Minuit` class is mostly pure-Python, but all located in a Cython file, so has access to C++ / MINUIT)

I think just a 1-2 page write up with the basic info and references would be very helpful to many people that come to MINUIT / iminuit.

Is anyone interested in leading or collaborating on this in the near future (weeks / months)?

[...]

PR #218, opened 2018-03-07
Citation

HDembinski:
Added a citation file, addressing issue #179 

Everyone, please check that your names are written correctly!

### DIFF
--- a/CITATION
+++ b/CITATION
@@ -6,7 +6,7 @@ Recommended BibTeX entries for the above citations are:
 
 @ONLINE{iminuit,
   author = {iminuit team},
-  title = {iminuit - A Python interface to MINUIT \cite{1975CoPhC..10..343J}},
+  title = {iminuit -- A Python interface to Minuit},
   year = {2012--},
   url = {https://github.com/iminuit/iminuit},
   note = {[Online; accessed 2018.03.05]}
--- a/CITATION
+++ b/CITATION
@@ -1,0 +1,29 @@
+If you use iminuit for work/research presented in a publication, we recommend and encourage to add the following two references:
+
+Piti Ongmongkolkul and Christoph Deil and Hans Dembinski and David Men\'endez Hurtado and Chris Burr and Andrew ZP Smith and Fabian Rost and Alex Pearce and Lukas Geiger and Omar Zapata, iminuit - A Python interface to MINUIT, 2012 --, https://github.com/iminuit/iminuit; accessed online <your current date>.
+
+The original Minuit paper.
+http://adsabs.harvard.edu/abs/1975CoPhC..10..343J
+
+Recommended BibTeX entries for the above citations are:
+
+@ONLINE{iminuit,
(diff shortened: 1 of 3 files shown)


CITATION
```diff
@@ -0,0 +1,29 @@
+If you use iminuit for work/research presented in a publication, we recommend and encourage to add the following two references:
+
+Piti Ongmongkolkul and Christoph Deil and Hans Dembinski and David Men\'endez Hurtado and Chris Burr and Andrew ZP Smith and Fabian Rost and Alex Pearce and Lukas Geiger and Omar Zapata, iminuit - A Python interface to MINUIT, 2012 --, https://github.com/iminuit/iminuit; accessed online <your current date>.
```
cdeil: Suggest to avoid the names here.
It'll already be painful to maintain the list of names in the bibtex entry below, but duplicating it here on this line just make it harder.

For presentations, I think we'll rather encourage people to put the logo / URL.

For papers, I think iminuit will usually appear in the acknowledgements, along with other sofware and services used.
How about suggesting a sentence to put, like they do e.g. in http://www.astropy.org/acknowledging.html ?

For us it could be:

> This research made use of MINUIT \cite{1975CoPhC..10..343J} via the iminuit \cite{iminuit} Python interface.

Or if you want to put iminuit first, it could be:

> This research made use of iminuit \cite{iminuit}, a Python interface to MINUIT \cite{1975CoPhC..10..343J}.

CITATION
```diff
@@ -0,0 +1,29 @@
...
+
+The original Minuit paper.
+http://adsabs.harvard.edu/abs/1975CoPhC..10..343J
+
+Recommended BibTeX entries for the above citations are:
+
+@ONLINE{iminuit,
+  author = {Piti Ongmongkolkul (@piti118) and Christoph Deil (@cdeil) and Hans Dembinski (@HDembinski) and David Men\'endez Hurtado (@Dapid) and Chris Burr (@chrisburr) and Andrew ZP Smith (@energynumbers) and Fabian Rost (@fabianrost84) and Alex Pearce (@alexpearce) and Lukas Geiger (@lgeiger) and Omar Zapata (@omazapa)},
```
cdeil: Is it a good idea to put the Github usernames in the author field?
Does this even render OK in LaTeX?
I think most people will not understand what this is, so I'm -1 to put it in the bibtex entry.

Adding a contributor list to our docs would be good, and there you could link to the Github handles.
This I find very convenience, I've done this e.g. for another project I work on: http://gammapy.org/team.html#contributors

HDembinski: Sounds good, that I did not remove the github handles was an oversight, not intentional. Should be skip the names and just put "iminuit collaboration"? And then list the names in a "contributors" section in the docs? I like this, because keeping this information in sync and up-to-date is a pain.

We want to have only one place where we put the contributors, it should be a "contributors" section or a CONTRIBUTORS file in the repo.

CITATION
HDembinski: Good point, I agree.

CITATION
cdeil: Yes, I would prefer "iminuit contributors" here, and only have to maintain a contributor list in one place, somewhere in the docs.

cdeil:
Looks good to me, I'm merging this now.

Of course, @piti118 or anyone: if you have a different suggestion, comments here or follow-up PRs suggesting changes are still welcome.

Issue #200, opened 2018-02-16
interface request: add np_vector, similar to np_matrix


HDembinski:
```iminuit.Minuit``` already provides a nice convenience interface to get the full covariance matrix after the fit: ```iminuit.Minuit.np_matrix()```. Unfortunately, there is no equivalent method for getting the parameter vector. For consistency reasons and because I personally need this all the time, I would like to see a method ```iminuit.Minuit.np_vector()``` that returns a vector of the fitted parameters in the same order as `np_matrix` is returned.

**Comment by cdeil** [comment]
I see that `Minuit.np.matrix` is basically a one-liner calling `np.array` on `Minuit.matrix`:
https://github.com/iminuit/iminuit/blob/fc5cc5d89f15921fbd78b27396eda22a5f99b487/iminuit/_libiminuit.pyx#L649
```
np_matrix = np.array(minuit.matrix())
```

I looked around a bit, and it seems no equivalent of `Minuit.matrix` exists, that exposes an ordered parameter list. One currently has to use this to turn the `Minuit.values` dict into a list with the parameters in the right order?
```
[m.values[name] for name in m.list_of_vary_param()]
```
Or did I just miss that property and it exists already?

@HDembinski - So basically what you're proposing is a property that returns this, no?
```
np.array(m.values[name] for name in m.list_of_vary_param())
```

---

In principle I'm +1 to add such a convenience(s).

But given that anything we add should remain there forever, I would prefer us to be careful / conservative with additions and to have someone review and propose a full addition of more Python-based properties (using tuples / lists / dicts, or instead changing to OrderedDict to avoid duplicating properties just to get sorted lists), and then which properties we really want to re-expose using Numpy arrays instead of just letting the caller pass an existing list or tuple into `np.array`. Note that there's different variants like "all parameters" or only the "varying parameters" or ... it would also be reasonable to add 10 new Numpy array convenience properties.

The alternative to adding conveniences is to improve the docs. Probably some mix of those two options would be idea.

@HDembinski @piti118 @alexpearce or anyone - what do you propose we add? Anyone willing to send a pull request (maybe wait for a few days to get opinions here)?

cdeil:
> For people who work with numpy a lot, it is simply more natural to get the fit result and the covariance matrix as numpy ndarrays. It scales better when you need to change the number of fitted parameters a lot, ...

[...]

PR #219, opened 2018-03-07
numpy getters

HDembinski:
Addresses issue #200 

The tuple-based attribute interface `values`, `errors`, `merrors`, and `covariance` is mirrored by numpy-friendly getters with the `np_` prefix. This adds a great deal of convenience for numpy users.

### DIFF
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -624,6 +624,16 @@ cdef class Minuit:
             a[i] = self.errors[k]
         return a
 
+    def np_merrors(self):
+        """Parameter errors in numpy array format."""
+        import numpy as np
+        # array format follows matplotlib conventions, see pyplot.errorbar
+        a = np.empty((2, len(self.parameters)), dtype=np.double)
+        for i, k in enumerate(self.parameters):
+            a[0, i] = -self.merrors[(k, -1.0)]
+            a[1, i] = self.merrors[(k, 1.0)]
+        return a
+
     def np_covariance(self):
         """Covariance matrix in numpy array format.
 
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -611,6 +611,35 @@ cdef class Minuit:
         matrix = self.matrix(correlation=correlation, skip_fixed=skip_fixed)
         return np.array(matrix, dtype=np.float64)
 
+    def np_values(self):
+        """Parameter values in numpy array format."""
+        import numpy as np
+        a = np.empty(len(self.parameters))
+        for i, k in enumerate(self.parameters):
+            a[i] = self.values[k]
+        return a
+
+    def np_errors(self):
+        """Parameter errors in numpy array format."""
+        import numpy as np
+        a = np.empty(len(self.parameters))
+        for i, k in enumerate(self.parameters):
+            a[i] = self.errors[k]
+        return a
+
+    def np_covariance(self):
+        """Covariance matrix in numpy array format.
+
+        Note that a ``numpy.ndarray`` is returned, not a ``numpy.matrix``
+        """
+        import numpy as np
+        n = len(self.parameters)
+        a = np.empty((n, n))
+        for i1, k1 in enumerate(self.parameters):
+            for i2, k2 in enumerate(self.parameters):
+                a[i1, i2] = self.covariance[(k1, k2)]
+        return a
+
     def is_fixed(self, vname):
         """Check if variable *vname* is (initially) fixed"""
         if vname not in self.parameters:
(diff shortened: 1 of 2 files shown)


cdeil:
👍 for this addition.

Some feedback:

Have you considered making those properties instead of methods? Especially given that `iminuit` uses properties for the equivalent info as Python types, that might make sense? If you think that we might ever want to add options here e.g. to sub-set to the free parameters only, then a method like you have now is the better choice, because it's extensible in a backward-compat way via keyword arguments.

Apart from that question, the implementation and tests look good to me.

Could you please be a bit more specific in the docstrings what exactly is returned?

I forgot: are the equivalent Python properties the same: always the full list of parameters, not just the free ones? If they behave differently, then this needs to be pointed out. If it's the same, I still think it's worth mentioning, or to add "this is equivalent to the info from <values/covariance>", i.e. point to the Python property.

For `np_merrors` it should also be clear from the docstring what the order of up / down is. You have a comment in the code, but that is not visible to users, they will always wonder. You could even be very explicit, and give a code example that shows that `m.np_merrors[0, :]` is the down error and `m.np_merrors[1, :]` is the up error (or the other way around, whatever is the correct way to get at those).

iminuit/_libiminuit.pyx
```diff
@@ -611,6 +611,42 @@ cdef class Minuit:
         matrix = self.matrix(correlation=correlation, skip_fixed=skip_fixed)
         return np.array(matrix, dtype=np.float64)
 
+    def np_values(self):
+        """Parameter values in numpy array format."""
+        import numpy as np
+        return np.asarray(self.args, dtype=np.double)
```
cdeil: Doesn't really matter, but `np.array` here instead of `np.asarray` is probably a tiny bit faster and two characters less. You know that you're calling on a tuple and need to convert to an array, whereas `np.asarray` does a lot of stuff to look at types and numpy array subclasses and ...

iminuit/_libiminuit.pyx
```diff
@@ -611,6 +611,42 @@ cdef class Minuit:
...
+        return a
+
+    def np_covariance(self):
+        """Covariance matrix in numpy array format.
+
+        Note that a ``numpy.ndarray`` is returned, not a ``numpy.matrix``
+        """
+        import numpy as np
+        n = len(self.parameters)
+        a = np.empty((n, n), dtype=np.double)
```
cdeil: Why not put `return self.np_matrix(skip_fixed=False)` here?
It's less code for us to maintain, and guarantees that `np.covariance` is consistent with `np.matrix` by design, whereas with the current PR I'd have to review the two implementations to see if they match.

iminuit/tests/test_iminuit.py
```diff
@@ -433,33 +433,68 @@ def f(x, y, z):
...
+        expected = [5.**0.5, 1.]
+        assert_allclose(actual, expected, atol=1e-8)
+        assert isinstance(actual, np.ndarray)
+
+    def test_np_merrors(self):
+        import numpy as np
+        actual = self.m.np_merrors()
+        expected = [[5.**0.5, 1.], [5.**0.5, 1.]]
+        assert_allclose(actual, expected, atol=1e-8)
+        assert isinstance(actual, np.ndarray)
```
cdeil: Can you please change this `isinstance` assert to an assert on `actual.shape`?
That tests a little bit more.
E.g. merrors is a 2-dim array, but values and errors should be 1-dim arrays, but from your current test would also pass if they were returned as 2-dim arrays with shape `(n_par, 1)`.
Looking at your implementation, they aren't, all is good.
But I think if it is easily possible to write tests that fully establish the behaviour without having to look at the implementation, that that is nice.

iminuit/_libiminuit.pyx
HDembinski: Good point, I wasn't aware that `array` is faster.

iminuit/_libiminuit.pyx
HDembinski: Because `skip_fixed=False` is raising a `RuntimeError` :) But you are right, I will fix `matrix()` and then do as you suggest, makes much more sense. Sorry, I forgot this, but we talked about it offline.

iminuit/tests/test_iminuit.py
HDembinski: Sure, good point.

cdeil:
@HDembinski - This looks good to me know, except for the docstring improvements I suggested in my last comment, expecially on `np_merrors` to say if up or down come first, but also generally for all to say that all parameters, including fixed ones are there.

cdeil:
@HDembinski - Thanks for improving the description. Especially for the complex case of `np_merrors` it's very good.

#200 and this PR have been open for a long time, and no one objected. So this is going in.

If anyone has concerns about these added methods, please speak up ASAP, or we shall keep them forever.

Issue #213, opened 2018-03-02
python setup.py test is broken


cdeil:
I see this error locally:
```
$ python setup.py test
running test
Traceback (most recent call last):
  File "setup.py", line 140, in <module>
    'build_ext': SmartBuildExt,
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/setuptools/command/test.py", line 216, in run
    cmd = ' '.join(self._argv)
TypeError: sequence item 2: expected str instance, bool found
```

I think this was there all the time, and we just didn't invocate the tests like that.

I think copy & pasting the latest code from here to update our setup.py would fix the issue:
https://docs.pytest.org/en/latest/goodpractices.html#manual-integration

Or we switch to this:
https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner

I haven't looked at the pros and cons of the different ways to run tests in a while, so I'm not sure.
@HDembinski - Any opinions?
If not, we could keep as-is and I just do the update in `setup.py` to fix this.

HDembinski:
@cdeil I looked into `setup.py` and apparently we went for 'manual-integration', it is just broken. I tried a few things to get it working again, and got some improvement, but then another failure. I don't really understand the error and how all this works. I am starting to find setuptools really annoying...

HDembinski:
My preferred solution would be to drop setup.py and use cmake to build this project. This would solve several of the issues I have with the current build. I don't know how that would work with pypi integration?

cdeil:
[...]

PR #221, opened 2018-03-07
fixed issue 213

HDembinski:
This fixes issue #213


cdeil:
@HDembinski - Thanks!

I note that our current CI isn't great:
https://github.com/iminuit/iminuit/blob/master/.travis.yml
It doesn't test this way of running the tests.
From a quick look, I don't even know where the code is imported from during test execution.
There's an in-place build and an install!?

The best thing to do is to install the test code with the package, because then users can also test their installed version. And then to also execute the tests like that in CI, i.e. focus on testing the installed version.

But those are larger changes. I'm merging this PR now.

PR #222, opened 2018-03-13
added missing doc rule; removed doc-show rule; make sphinx find local build

HDembinski:
A "doc" rule was advertised in the "help", but not actually implemented. Now we have one, and it only rebuild the html docu if the source files changed.

I took the liberty to remove the "doc-show" rule, which only works on Mac.

Finally, I added a bit of code to doc/conf.py, which enables sphinx to find the locally build iminuit module. When you are developing, that is usually what you want. Before, you had to install iminuit at least user-wide just to generate the docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ help:
 	@echo '     coverage         Run tests and write coverage report'
 	@echo '     cython           Compile cython files'
 	@echo '     doc              Run Sphinx to generate HTML docs'
-	@echo '     doc-show         Open local HTML docs in browser'
 	@echo ''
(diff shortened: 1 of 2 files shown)


cdeil:
Sounds good to me. Thanks!

cdeil:
@HDembinski - Can you please check http://readthedocs.org/projects/iminuit/builds/6896185/ and the updated online docs after 10 min or so?

I'm not sure if this PR has the potential to change or break something about the docs build on RTD. Also, sometimes one needs to "wipe" the build environment there to get a clean build after touching the build or docs build.

PR #224, opened 2018-03-17
fix: detect and handle functions with variable length args

HDembinski:
`describe` did not fail on functions with variable length args, while it should. Since it did not raise an error, Minuit later fails with an obscure error message when you try to call `migrad`.

Now an error is correctly raised by `describe` and the cases are tested.

### DIFF
--- a/iminuit/util.py
+++ b/iminuit/util.py
@@ -100,13 +100,12 @@ def arguments_from_funccode(f):
     """
     fc = fc_or_c(f)
     vnames = fc.co_varnames
+    nargs = fc.co_argcount
     # bound method and fake function will be None
-    if is_bound(f):
-        # bound method dock off self
-        return list(vnames[1:fc.co_argcount])
-    else:
-        # unbound and fakefunc
-        return list(vnames[:fc.co_argcount])
+    args = vnames[1:nargs] if is_bound(f) else vnames[:nargs]
+    if not args:
+        raise RuntimeError('Function has variable number of arguments')
+    return list(args)
 
 
 def arguments_from_call_funccode(f):
--- a/iminuit/util.py
+++ b/iminuit/util.py
@@ -102,7 +102,7 @@ def arguments_from_funccode(f):
     vnames = fc.co_varnames
     nargs = fc.co_argcount
     # bound method and fake function will be None
-    args = vnames[1:nargs] if is_bound(f) else vnames[:nargs]
+    args = vnames[1 if is_bound(f) else 0:nargs]
     if not args:
         raise RuntimeError('Function has variable number of arguments')
     return list(args)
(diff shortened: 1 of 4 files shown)


cdeil:
@piti118 - Could you please have a look?

(I'm not very familiar with the function inspect stuff)

HDembinski:
All this patch does is to make sure that the following code raises a sensible exception, while it was silently passing before
```
from iminuit import Minuit
m = Minuit(lambda *args: 0)
# should fail, because there is no way to know the names and number of parameters
```

iminuit/util.py
```diff
@@ -100,13 +100,12 @@ def arguments_from_funccode(f):
...
     # bound method and fake function will be None
-    if is_bound(f):
-        # bound method dock off self
-        return list(vnames[1:fc.co_argcount])
-    else:
-        # unbound and fakefunc
-        return list(vnames[:fc.co_argcount])
+    args = vnames[1 if is_bound(f) else 0:nargs]
+    if not args:
+        raise RuntimeError('Function has variable number of arguments')
```
cdeil: I would put `ValueError`. It's a classical case of a user passing in an invalid value (i.e. a function that can't be processed).

HDembinski: Sounds good, but then we should change all the errors. Please have a look into the file, all the functions there throw RuntimeError. I followed that lead instead of throwing a more specific error. In fact, the error type does not matter, because any error thrown in one of these utility functions is converted into a TypeError in `better_arg_spec`.

cdeil: I wasn't aware. Then let's keep as-is.

HDembinski: Perhaps we can improve the code in utils.py at some later point. Raising errors, then capturing them to print a message and then throwing another error again looks a bit inelegant.

cdeil: Catching a RunTimeError is terrible. It can probably occur from, well, runtime issues.
It should never be caught and masked.

The pattern to catch errors is not bad, but for those cases, a custom exception that's internal to the library should be created.

HDembinski: The raised errors here are used as an internal control structure. I think that's not a good use for errors. I would not catch errors internally to transform them and give other errors to the user, that seems superfluous. Raising and catching errors is very slow in C++, probably less so in Python (because everything is slow in Python), but it should not be used for an internal control structure. I would either hand the original error to the user or not raise errors but indicate internal failures in some other way, e.g. by returning None. Errors are there to propagate to the user.

HDembinski: Relevent SO discussion: https://stackoverflow.com/questions/7274310/python-using-exceptions-for-control-flow-considered-bad

cdeil:
@HDembinski - Let's continue the discussion about larger changes to the error handling later.

If I understand correctly, you didn't want to do them here, and anyways they would likely be backward-incompat, so best to put in a separate PR.

Sounds like the proper solution is to remove the catch and re-raise, and to directly raise TypeError.
Why was the catch and re-raise done? To print something? Libraries should never print.
If it was to emit a log message, it might have been a good idea and needs a closer look.

iminuit/util.py
HDembinski: So we agree.

PR #225, opened 2018-03-17
Add minimize function with scipy.optimize.minimize API

HDembinski:
Alternative interface for Minuit, full imitation of scipy.optimize.minimize. It is used like this
```
from iminuit import minimize

def f(x): return x[0]**2 + x[1]**2
result = minimize(f, (1,1))
# result is of type scipy.optimize.OptimizeResult and properly filled
```


iminuit/minimize.py
```diff
@@ -0,0 +1,109 @@
...
+             bounds=None, constraints=(),
+             tol=None,
+             callback=None, options=None):
+    """Imitates the interface of the SciPy method of the same name in scipy.optimize.
+
+    For argument description, see scipy.optimize.minimize.
+
+    The only supported method is 'Migrad'.
+    """
+    from numpy import array
```
cdeil: Please put as `import numpy as np`.
IMO makes it clearer to read in the code below ... `array` is also a common variable name.

iminuit/tests/test_minimize.py
```diff
@@ -0,0 +1,51 @@
...
+
+def grad(x, *args):
+    return np.array((2 * x[0], 2 * (x[1] - 1), 2 * (x[2] - 2)))
+
+
+def test_simple():
+    result = minimize(func, (1, 1, 1))
+    assert_allclose(result.x, (0, 1, 2), atol=1e-8)
+    assert_allclose(result.fun, 1)
+    assert(result.nfev > 0)
```
cdeil: assert is a statement, not a function call. Please remove the `()` everywhere you have assert statements.

iminuit/minimize.py
HDembinski: I have to disagree for several reasons. Firstly, calling `array` repeatedly is faster than `np.array`, because the latter makes an extra map lookup. We are talking about implementation code in a short function, so it is not a problem that `array` is a common variable name.
As a compromise, I could this import `from numpy import array as np_array`.

iminuit/tests/test_minimize.py
HDembinski: My bad, C++ idioms are leaking in.

HDembinski:
I agree with your suggestions and remarks. We need to support such numpy-array-based functions on a broader level in iminuit. I have some ideas on how to implement that in a proper way. The changes would be orthogonal to this PR.

The nice thing about this PR is that you can now easily switch in some script that uses `minimize` from `scipy.optimize` to `minimize` from `iminuit`. That makes it effortless for people to try it, and that's a selling point that could gain us more users. Once they tried it and like it, they should use the "real" interface, though. It is also handy for benchmark comparisons with `scipy`.

HDembinski:
It is a good idea to keep the minuit object around in the OptimizationResult dict. Let's do that.

iminuit/minimize.py
HDembinski: Proof
```
In [6]: timeit.timeit("a =np.array(0)", setup="import numpy as np")
Out[6]: 0.3645598888397217

In [7]: timeit.timeit("a =np_array(0)", setup="from numpy import array as np_arr
   ...: ay")
Out[7]: 0.3079559803009033
```

cdeil:
This looks good to me. Suggest to add this to iminuit. Add small high-level docs page in this PR or later.

HDembinski:
Somehow I messed up the patch history. I will make a new PR and merge this instead.

PR #228, opened 2018-04-17
Support function with sequence argument (full numpy support)

HDembinski:
After adding numpy support on the output side, this series of patches adds full support for functions with sequence arguments, including support for numpy's ufuncs. This is achieved seamlessly by enhancing the behavior of `Minuit.__init__` in a backward-compatible way.

Minuit accepts functions with the signature `def func(x): ...` now, where x is a sequence. This works even if numpy support is not compiled in. To tell Minuit that x is not a number but a sequence, simply pass a sequence in the keyword arguments x, error_x, limit_x, fix_x or similar. Example:

m = Minuit(func, x=(1, 2))

This is the new bit in the interface, that you can pass sequences in addition to numbers. When numpy support is compiled in (compiled numpy support is also new), then x is internally passed to func as a numpy array. When numpy is not available, x is passed as a tuple to func.

For convenience, also a named constructor was added, which makes this a bit easier. It enforces passing a starting sequence as a positional argument. It is not necessary to name it.
m = Minuit.from_array_func(func, (1,2)) # (1,2) is the starting sequence


HDembinski:
Something is wrong with Travis, I am trying to close this and reopen it.

cdeil:
Just for the record: I think this was continued in #229

Issue #185, opened 2016-04-03
grad_fcn with parameters given in an array 


rodiram:
Adding the gradient option in iminuit.Minuit speeds-up a lot the minimization when numerous variables are used. Thank you for this recent improvement.
From what I've understood, the variable values at which the gradient will be calculated must be given explicitly in the gradient function: 
def func4_grad(x, y, z):
...
this is not possible when fitting lot's of variables (cases for which the gradient input is needed).
Is there a way to use an array as argument to the gradient function ? I tried but it did not work, I could not find the corresponding line code in the sources files. Something like:
def func4_grad(double[::1] my_variables):
...
Thanks !

rodiram:
I see that in PythonGradientFCN.h the grad function is called with a vector in Gradient:
Gradient(const std::vector<double>& x)
PyObject* tuple = vector2tuple(x);
PyObject* result = PyObject_Call(gradfcn,tuple,NULL);
..
But I can not find the link between this function and the grad_fcn function of _libiminuit.pyx which is provided by the user.
Can someone points the lines where it is done ?

rodiram:
Is iminuit still maintained ? If not it would be useful to mention it somewhere since it is an important parameter when one chooses an option or another. Thank you anyway for this work.

cdeil:
@HDembinski - Assigning to you and the v1.3 milestone.

Does this require further code changes? Or "just" docs improvements to explain how to use functions with gradients or the new array interface that you added recently?

HDembinski:
This feature works now with the master, no further code changes needed. I just added an explicit test for this case. All this will be in the docs when we describe the new feature of using Minuit with array-based functions.

PR #229, opened 2018-04-18
Numpy support

HDembinski:
After adding numpy support on the output side, this series of patches adds full support for functions with sequence arguments, including support for numpy's ufuncs. This is achieved seamlessly by enhancing the behavior of `Minuit.__init__` in a backward-compatible way.

Minuit accepts functions with the signature `def func(x): ...` now, where x is a sequence. This works even if numpy support is not compiled in. To tell Minuit that x is not a number but a sequence, simply pass a sequence in the keyword arguments x, error_x, limit_x, fix_x or similar. Example:

m = Minuit(func, x=(1, 2))

This is the new bit in the interface, that you can pass sequences in addition to numbers. When numpy support is compiled in (compiled numpy support is also new), then x is internally passed to func as a numpy array. When numpy is not available, x is passed as a tuple to func.

For convenience, also a named constructor was added, which makes this a bit easier. It enforces passing a starting sequence as a positional argument. It is not necessary to name it.
m = Minuit.from_array_func(func, (1,2)) # (1,2) is the starting sequence


HDembinski:
Addresses #185 and is a more powerful alternative to #190

iminuit/_libiminuit.pyx
```diff
@@ -1424,3 +1590,81 @@ cdef class Minuit:
                 raise RuntimeError(
                     ('Cannot understand keyword %s. May be a typo?\n'
                      'The parameters are %r') % (k, parameters))
+
+    IF HAVE_NUMPY:
```
cdeil: Is it a good idea to put this `IF HAVE_NUMPY` before the `def np_matrix`?
I think it will just lead to users getting a cryptic `AttributeError` instead of the normal `ImportError` when they run a script that call `np_matrix`, no?
Suggest to change this back.

Btw, why is it `IF` and not `if`?

HDembinski: IF is a compile-time if, not a run-time if. You can look it up in the Cython documentation. The methods np_* are now only added to the Minuit object when numpy was available during the compilation of the module.

That these methods were there even if numpy as not available was a bug, now it is consistent.

HDembinski:
1. There is a separate create method and init was enhanced. I thought about this and this is the more intuitive solution for the user. It is also easier to implement with less code duplication. Now the create method is a very thin method, and the meat that makes this work is in the init method, which allows a lot of code reuse.

2. This is not about performance, but about convenience. Why add numpy interface in this way? Well, everyone wants to use ufuncs with Minuit and it is a pain that we do not support that. I know several people who want this, myself included. If you read my explanation closely, you will see that this patch is not only about adding numpy support, but about adding support for FCNs that accept the parameters as a sequence argument, exactly a ufunc would. This is useful even if you don't use numpy (but who seriously does scientific computing without numpy).

HDembinski:
Secondly, we don't know how people use Minuit and we should prepare for the worst. Some people will use Minuit with functions that accept 10+ parameters.

HDembinski:
Finally, this is not a change. This is an enhancement! Every piece of code that worked before still works. This patch merely allows users to do more with Minuit.

iminuit/_libiminuit.pyx
cdeil: All project I know use the "delayed import" in the function / method for optional dependencies.
Not the technique with the compile-time IF you introduce here, to hide the methods if the dependency is not available.
But ImportError is better than AttributeError if Numpy is not available, and also to "see" that the method exists even if no Numpy is installed, so that I can just install it, is the better user experience, no?

HDembinski: The `np_*` methods should not be there in the first place when numpy is not available on your system. If they are there, this is an interface bug. They should not show up when you do `help(iminuit.Minuit)` in the interpreter.

The other projects you mention do it this way, because they don't know how to do a better job. I spend some hours on how to do this properly. The IF statement seems to be a recent addition to Cython. It probably was not available when these other projects were written.

HDembinski: I also challenge the statement that ImportError is better than AttributeError. What is the error you should naively expect when you try to call a method that logically should not be there? You should get an AttributeError.

cdeil: Come one, that's just a matter of preference. There is no right or wrong here. Most other projects do it the way we had it and that I prefer. It's technically trivial to do it the way you do it for any Python project.

But that's a minor change here. Maybe we table the discussion or wait for someone else to chime in. And hopefully someone will have time / expertise to review the rest of the PR.

HDembinski: No, it is not a matter of preference. In design, there is right and wrong. Once we agree on the criteria by which we judge a design, we can draw an objective conclusion. The subjectivity is  in the criteria by which to judge a design.

In the case of Python, the design criteria have been informally written down in form the Zen of Python. I like Python because of these criteria, for me they are intuitively obvious. Others have to accept them as authority.
"Beautiful is better than ugly."
"Explicit is better than implicit."
"Simple is better than complex."
"Readability counts."

In my own words: things should be obvious in Python, no false leads. If your class has a method that consistently fails whenever you call it, that method should not be there in the first place. It should not appear in the documentation, when you use the online help. Announcing methods in the online help that do not work is bad design. It is ugly, it is a false lead, the error is implicit, and it reduces readability of the documentation, because it is a distraction.

HDembinski: Yes, let's talk in person about these things.

Still, here is the thing that I think you are not seeing yet. In the new version of iminuit, you have to decide at compile-time whether Numpy support is enabled or not. The new default will be Numpy Support ON, with Numpy Support OFF as an option. This is an acceptable increase in dependencies, because everyone uses Numpy anyway. Once Numpy Support is a compile-time decision and you turn it off, it makes no sense to have methods that require numpy to be present. It is more consistent to remove these methods then, because they are useless clutter.

Imagine you had a door with two door handles, one red and one blue. The door would only operate with the red handle, while the blue one would give you a mild electric shock. No designer would keep the blue handle around.

iminuit/_libiminuit.pyx
```diff
@@ -192,10 +195,39 @@ cdef class Minuit:
 
         **Arguments:**
 
-            - **fcn**: the function to be optimized. Minuit automagically finds
-              parameters names. More information about how
-              Minuit detects function signature can be found in
-              :ref:`function-sig-label`
+            **fcn**, the function to be optimized, is the only required argument.
+
+            Two kinds of function signatures are understood.
```
cdeil: @HDembinski  and I had a long discussion on this PR and specifically the questions whether to support a second type of callback function (that takes a sequence) and whether to add a second create classmethod to support the use case currently covered by forced_parameters better.

In the end, it still remains a complex decision, and a matter of preference whether convenience / removed boilerplate for certain use cases is important enough to expand the API.

For me, the bottom line is that my preference would be to mostly keep things as is, concretely:
- add documentation how to do common things with the current API (e.g. using `def f(*pars): pars = np.array(pars)` and explain to users what `*pars` is), not add a second way to do this as proposed here via `def f(pars)` where pars is directly an array
- OK to add a classmethod that helps to init this use case, which basically prepares a `kwargs` dict and calls `init(**kwargs, forced_parameters=True)` but makes this more convenient.

My argument is that there's a hidden cost to adding new features and a second way to do the same thing, because then we need to explain both ways in the docs and users aren't better off really if half of them use one way and the other the other way, eventually they have to learn and deal with both ways.
Brian Granger made similar arguments much better than I could here: https://youtu.be/OrpPDkZef5I

But I admit this new feature is nice, and decided to leave the decision on this to @HDembinski .
That is, unless @piti118 - if you chime in here, it would be very valuable.

HDembinski:
@cdeil I implemented the changes that we discussed. The `__init__` method is now mostly restored to its original form, the additional complexity removed. I also removed other changes which were not related to the feature of this branch. Do you want to have another look or can I merge this?

setup.py
```diff
@@ -56,10 +59,40 @@ def run_tests(self):
         sys.exit(errno)
 
 
+# delay import of numpy until it is installed
```
cdeil: It's not obvious to me what is going on here. Could you please leave a longer comment here or in the docs?
Numpy is now a required dependency?
And this delayed import is needed because `setup.py` is executed by `pip` before `numpy` as a dependency is installed?

setup.py
```diff
@@ -56,10 +59,40 @@ def run_tests(self):
...
+def lazy_numpy_get_include():
+    try:
+        import numpy
+        return [numpy.get_include()]
+    except:
```
cdeil: except ImportError?

iminuit/_libiminuit.pyx
```diff
@@ -328,25 +372,81 @@ cdef class Minuit:
...
+            - Upper limit: (None, x), (-inf, x) [replace x with a number]
+
+            **fix**: Optional sequence of boolean values. Whether to fix a
+            parameter to the starting value.
+
+            **name**: Optional sequence of parameter names. If names are not
+            specified, the parameters are called x0, ..., xN.
+
+            All other keywords are forwarded to :class:`Minuit`, see
+            its documentation.
```
cdeil: Working simple example (probably in the high-level docs and only linked to from here) is gold for users.

setup.py
HDembinski: Yes, Numpy is now a required dependency. This one of things we discussed and you preferred that over the more complex code I had before, which kept numpy optional. Do you remember?

HDembinski: It is  not clear to me whether this code here is strictly necessary. The general idea is to delay the import of numpy until it is installed by pip as part of the requirements if you don't already have it. I have no clear understanding how setup.py interplays with pip. I just copied this more or less. There is not much more to comment here.

setup.py
HDembinski: OK

iminuit/_libiminuit.pyx
HDembinski: Right, but let's work on the docs in a separate PR, since there is more to do than adding an example for this.

setup.py
cdeil: I remember the discussion, but to be honest not the important detail whether compile-time linking against Numpy is needed / useful or not.

Is compile-time linking required to support the second cost function type you are adding here? I think this is the case, even if internally in iminuit everything still goes through Python types, i.e. there is no performance benefit to introduce this.

Ideally there would be a few lines at
http://iminuit.readthedocs.io/en/latest/installation.html
describing this, and then maybe a link to that description from the `setup.py` file.
Or you decide which docs are better suited as code comments or for the docs, so that the design and build is clear to users and devs.

HDembinski: > I remember the discussion, but to be honest not the important detail whether compile-time > linking against Numpy is needed / useful or not.
> Is compile-time linking required to support the second cost function type you are adding
> here? I think this is the case, even if internally in iminuit everything still goes through
> Python types, i.e. there is no performance benefit to introduce this.

Yes, it must be exactly like this. We discussed all of that and I will not discuss it again here, it is tiring to repeat the same technical arguments over and over.

I will remove the lazy numpy import if that makes you happy.

cdeil: I'm just making small suggestions to add a few lines documenting how the new solutions work. Up to you to take them or just merge.

Issue #232, opened 2018-06-08
Why console printout `import re`?


cdeil:
With iminuit master and the example from the docs page I get this output.
Why is there a printout of `import re`!?

@HDembinski - Do you know / see the same? Or should I investigate?

```
In [3]: %paste
from iminuit import Minuit
def f(x, y, z):
    return (x - 2) ** 2 + (y - 3) ** 2 + (z - 4) ** 2
m = Minuit(f)
m.migrad()
print(m.values)  # {'x': 2,'y': 3,'z': 4}
print(m.errors)  # {'x': 1,'y': 1,'z': 1}

## -- End pasted text --
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: errordef is not given. Default to 1.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter x does not have initial value. Assume 0.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter x is floating but does not have initial step size. Assume 1.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter y does not have initial value. Assume 0.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter y is floating but does not have initial step size. Assume 1.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter z does not have initial value. Assume 0.
  import re
/Users/deil/software/anaconda3/envs/iminuit/bin/ipython:4: InitialParamWarning: Parameter z is floating but does not have initial step size. Assume 1.
  import re
{'x': 2.0000000000047327, 'y': 3.000000000007099, 'z': 4.000000000009465}
{'x': 1.0000000000000409, 'y': 1.0000000000002762, 'z': 0.9999999999998322}
```

HDembinski:
Hi, I can confirm this with IPython, but it does not appear in the normal Python interpreter. Also, it does not appear when the function is a lambda, only when it is a real function.

HDembinski:
Minimal code:
```
from iminuit import Minuit
def f(x): return x** 2
Minuit(f)
```

HDembinski:
[...]

PR #233, opened 2018-06-14
fix warning format bug with custom warning formatter

HDembinski:
Fixes issue #232


HDembinski:
Closing this PR, because it creates side effects. Something like this

https://stackoverflow.com/questions/2647128/emitting-cythonic-warnings

should be a better solution as pointed out by @cdeil.

PR #235, opened 2018-06-20
Minimize

HDembinski:
I technically could not merge #225. This is a replacement.

The minimize function now uses the new Minuit interface for numpy-array functions. It also adds the possibility to query how often the gradient function is called by MIGRAD. This is needed by the minimize result and it is a good feature to have.


HDembinski:
I am merging this, since the original PR was already approved.

Issue #234, opened 2018-06-20
Installation fails without cython


raymondEhlers:
Thanks for your hard work on this excellent package! Attempting to install the master without `cython` fails with the following output:

```
(testEnv) user@MBP ~/code/test/minuitTests/iminuit $ pip install git+git://github.com/iminuit/iminuit.git#egg=iminuit
Obtaining file:///Users/user/code/test/minuitTests/iminuit
Requirement already satisfied: setuptools in /Users/user/code/test/minuitTests/testEnv/lib/python3.6/site-packages (from iminuit==1.2) (39.0.1)
Requirement already satisfied: numpy in /Users/user/code/test/minuitTests/testEnv/lib/python3.6/site-packages (from iminuit==1.2) (1.14.5)
Installing collected packages: iminuit
  Running setup.py develop for iminuit
    Complete output from command /Users/user/code/test/minuitTests/testEnv/bin/python3 -c "import setuptools, tokenize;__file__='/Users/user/code/test/minuitTests/iminuit/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps:
    Cython is not available ... using pre-generated C file.
    running develop
    running egg_info
    writing iminuit.egg-info/PKG-INFO
    writing dependency_links to iminuit.egg-info/dependency_links.txt
    writing requirements to iminuit.egg-info/requires.txt
    writing top-level names to iminuit.egg-info/top_level.txt
    reading manifest file 'iminuit.egg-info/SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    warning: no files found matching '*.cpp' under directory 'iminuit'
    writing manifest file 'iminuit.egg-info/SOURCES.txt'
    running build_ext
    building 'iminuit._libiminuit' extension
    creating build
    creating build/temp.macosx-10.13-x86_64-3.6
    creating build/temp.macosx-10.13-x86_64-3.6/Users
    creating build/temp.macosx-10.13-x86_64-3.6/Users/user
    creating build/temp.macosx-10.13-x86_64-3.6/Users/user/code
    creating build/temp.macosx-10.13-x86_64-3.6/Users/user/code/test
    creating build/temp.macosx-10.13-x86_64-3.6/Users/user/code/test/minuitTests
    creating build/temp.macosx-10.13-x86_64-3.6/Users/user/code/test/minuitTests/iminuit
[...]

PR #236, opened 2018-06-20
Fix cython dep

HDembinski:
This is a fix for #234. Instead of just throwing an error, display a helpful error message if cython is not available. Also fixes installation of iminuit without Cython if the pregenerated cpp file is actually there. This used to work and got broken. We need this for the next release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,8 @@ matrix:
 before_install:
   - pip install setuptools numpy cython
   - if [[ $BUILD == NO_CYTHON ]]; then
-      python setup.py sdist # use this to generate cpp files from pyx files
-      pip uninstall cython
+      python setup.py sdist; # use this to generate cpp files from pyx files
+      pip uninstall cython;
     fi
   - if [[ $BUILD != NO_EXTRA_DEPS ]]; then pip install IPython matplotlib; fi
   - if [[ $BUILD == DOCS ]]; then pip install matplotlib sphinx sphinx_rtd_theme; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,10 @@ matrix:
 
 before_install:
   - pip install setuptools Cython
+  - if [[ $BUILD != NO_CYTHON ]]; then
+      pip install sdist;
+      pip uninstall cython;
+    fi
   - if [[ $BUILD != NO_EXTRA_DEPS ]]; then pip install IPython numpy matplotlib; fi
   - if [[ $BUILD == DOCS ]]; then pip install matplotlib sphinx sphinx_rtd_theme; fi
   - if [[ $BUILD == TEST ]] || [[ $BUILD == COVERAGE ]]; then pip install pytest pytest-cov numpy; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
   - pip install setuptools Cython
   - if [[ $BUILD == NO_CYTHON ]]; then
       python setup.py sdist;
-      pip uninstall cython;
+      pip uninstall -y cython;
     fi
   - if [[ $BUILD != NO_EXTRA_DEPS ]]; then pip install IPython numpy matplotlib; fi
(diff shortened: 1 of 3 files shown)


cdeil:
I wouldn't have touched this, but if you put this in, OK, the change looks good.

IMO more important would be to give better instructions at
https://iminuit.readthedocs.io/en/latest/installation.html
specifically to split out the dev stuff (Cython / sphinx / pytest / build from source) into a separate page and there to explicitly give the command that Cython needs to be installed first.
@HDembinski - Should I make a PR for that or had you plans to tend to the docs also?

One more thought: what about the case where someone has Cython but installs a stable version?
There we currently re-run Cython, right?
Is that a good idea?
I guess the question is whether Cython or C++ compilers are more stable over the years.

cdeil:
Actually, I'm not 100% sure this is a good change, suggest to not put it in and stick to:
http://docs.cython.org/en/latest/src/reference/compilation.html#distributing-cython-modules

Note that in Python tools like pip or conda or build tools e.g. on PyPI or other packages need to run `python setup.py` just to get a version number and dependencies listed there. They can't know that Cython is needed to run `python setup.py`. So with this change, some people might get a `SystemError` for `python setup.py egg_info` where previously they got a version number?

Things are improving in Python (see e.g. https://www.python.org/dev/peps/pep-0517/ or 
 https://github.com/pypa/pipfile), but that's super new and I think relies on people having the latest pip, so I don't think we should use it yet here.

cdeil:
If you want to put this in: Maybe print something instead of raising a SystemError to be on the safe side?

HDembinski:
@cdeil I don't understand. A SystemExit prints the informative message and exists.

cdeil:
@HDembinski - A bit related to the discussion here, and FYI while you're touching `setup.py`.

There is a bit of an advantage to have `python setup.py` not error out if dependencies aren't installed yet:
```
$ python setup.py egg_info
```
E.g. `pip` will run `python setup.py` to figure out dependencies, so this could be problematic:
```
$ python setup.py egg_info
Cython is not available ... using pre-generated C file.
  File "setup.py", line 101, in <module>
    import numpy
ModuleNotFoundError: No module named 'numpy'
```

If you look at the `setup.py` for `scipy` or `scikit-learn` or any other well-done project, you'll find that there 
```
python setup.py egg_info
```
runs fine without any dependencies installed (like Numpy or Cython for us here).

Astropy even has a build to test that:
https://github.com/astropy/astropy/blob/666002ec20a9693d4f7742f18e0973ff76cba44b/.travis.yml#L73

HDembinski:
@cdeil I think this can be merged now. Do you want to test this some more?

.travis.yml
```diff
@@ -45,4 +46,4 @@ install:
 script:
   - if [[ $BUILD == TEST ]]; then make test; fi
   - if [[ $BUILD == COVERAGE ]]; then make coverage; fi
-  - if [[ $BUILD == DOCS ]]; then cd doc; sphinx-build -W -a -E -b html -d _build/doctrees . _build/html; fi
+  - if [[ $BUILD == DOCS ]]; then make doc; fi
```
cdeil: I think the `-W` makes the docs build fail if there is some problem, but `make doc` doesn't.
It's useful to catch docs build issues that way.
You could add a `make doc-ci` or find some other way to keep the CI behaviour if you agree.

setup.py
```diff
@@ -142,7 +152,7 @@ def get_version():
     extras_require={
         'all': ['ipython', 'matplotlib'],
     },
-    tests_require=['pytest', 'pytest-cov', 'numpy', 'scipy'],
+    tests_require=['pytest', 'pytest-cov', 'numpy', 'scipy', 'matplotlib', 'IPython<6'],
```
cdeil: Why pin to ipython < 6? Do we not want to support the latest stable ipython?
If this is really needed, suggest to at least leave a comment why this is done.

cdeil:
Looks good to me. Thanks!

PR #238, opened 2018-06-20
Major docs update

cdeil:
This is a major docs update.

The user install isntructions are now much simpler, and several other pages have been added.

screen shot 2018-06-20 at 16 06 54

@HDembinski - Please have a look.

I would like to merge this later today, and then continue on more docs improvements in follow-up PRs.
One thing I had planned to do here is to rewrite the dev docs page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ htmlcov
 .eggs
 env
 coverage.xml
+.pytest_cache
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 build
 *.pyc
 .idea
+.vscode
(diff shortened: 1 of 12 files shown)


HDembinski:
Looks good. I think reading the diff makes not much sense in this case. I volunteer to improve the About page in a next PR. I also would like to write on the Tutorial section.

cdeil:
I'm merging this now.

There's plenty of things left to improve with the docs, but I think it's easier to continue via new PRs, and to see the docs online, because the diff here is already quite large.

@HDembinski - Feel free to open issues and assign to me any time.

PR #158, opened 2018-06-20
Add Cython to coverage report

cdeil:
Now that coverage 4.0 and Cython 0.23 is out it should be easy to get coverage reports for Cython files:

http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html

This would be very nice to have for iminuit, because the core code is 1300 lines of Cython in `iminuit/_libiminuit.pyx`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,6 @@
 [run]
 omit = */py23_compat.py
+plugins = Cython.Coverage
 [report]
 exclude_lines =
     pragma: no cover
@@ -9,4 +10,3 @@ exclude_lines =
     __repr__
     __getitem__
     raise RuntimeError
-
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -1,3 +1,5 @@
+# cython: linetrace=True, binding=True
+# distutils: define_macros=CYTHON_TRACE=1
 # cython: embedsignature=True, c_string_type=str, c_string_encoding=ascii
 # distutils: language = c++
 """IPython Minuit class definition.


cdeil:
Attached a commit that yields a Cython coverage report.

I followed the suggestions here:
- http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html
- https://medium.com/@dfdeshom/better-test-coverage-workflow-for-cython-modules-631615eb197a

I don't think we should merge this, at least not before we have a better understanding how this affects the generated Cython.

Overall I have the impression coverage reporting is still very much work in progress for Cython:
https://github.com/cython/cython/search?q=coverage&type=Issues
https://github.com/cython/cython/issues/1461
Maybe we should just wait a year or two.

I'll open a few small issues making suggestions for test additions based on the local coverage report though.

HDembinski:
Where is the coverage report to see?

cdeil:
I didn't post it online. If you want it, please check out this branch and do
```
$ make coverage
$ open htmlcov/index.htm
```

HDembinski:
@cdeil This is great. Coverage reports, as we both know, are great to find dead code and to gain confidence that the project is fully tested.

HDembinski:
I am very much in favor of adding this. I don't know how to check for bad side effects that you worry about... if the tests run fine, everything should be ok, right?

cdeil:
The links above discuss the effect of 
```
# cython: linetrace=True, binding=True
# distutils: define_macros=CYTHON_TRACE=1
```

My understanding (more of a guess) is that `binding=True` is a good change we can safely make.

For `linetrace=True` there's probably a C code size increase, and for `distutils: define_macros=CYTHON_TRACE=1` there's probably a runtime slowdown.

So what we need is probably changes to `Makefile`, `setup.py` and dev docs to only keep the "production" cython / build as-is, and to add a special "coverage" build.

I could maybe do this, but really it's not clear to me, so @HDembinski - if you think you could add this relatively quickly, please go ahead. Otherwise I'll try to do it on Monday.

cdeil:
I'm moving this to the `whishlist` milestone; I don't have time this week, and generally my preference is to not touch `setup.py` or Cython options.

Developers that want a coverage report can get it locally for any branch within less than a minute just by copy & pasting these commands:
```
git cherry-pick -n 204cf55afc75aadc356ffddafdd6f648509766bd
git reset HEAD
make clean && make coverage
open htmlcov/index.html 
```

HDembinski:
Implemented in develop without relying on Cython

PR #163, opened 2018-06-20
Implement or remove get_initial_param_state

cdeil:
This is currently exposed by the API, but not yet implemented:

``` python
>>> from iminuit import Minuit
>>> def f(x, y, z):
...         return (x - 2) ** 2 + (y - 3) ** 2 + (z - 4) ** 2
... 
>>> m = Minuit(f, pedantic=False)
>>> m.get_param_states()
  File "<stdin>", line 1, in <module>
  File "iminuit/_libiminuit.pyx", line 718, in iminuit._libiminuit.Minuit.get_param_states (iminuit/_libiminuit.cpp:10274)
    if self.last_upst is NULL:
  File "iminuit/_libiminuit.pyx", line 732, in iminuit._libiminuit.Minuit.get_initial_param_state (iminuit/_libiminuit.cpp:10454)
    """Initial setting in form of MinuitParameter Struct"""
NotImplementedError
```

@piti118 – Can you fix this or comment what should be done? (I'm still not familiar with the Minuit code base ... slowly learning)

### DIFF
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -957,10 +957,6 @@ cdef class Minuit:
         """Dictionary of varname-> MinosError Struct"""
         return self.merrors_struct
 
-    def get_initial_param_state(self):
-        """Initial setting in form of MinuitParameter Struct"""
-        raise NotImplementedError
-
     def get_num_call_fcn(self):
         """Total number of calls to FCN (not just the last operation)"""
         cdef IMinuitMixinPtr ptr = dynamic_cast[IMinuitMixinPtr](self.pyfcn)


cdeil:
Moving this to the v1.3 milestone.

@HDembinski - Could you please make a decision here and on #230?

In case removing the `get_initial_param_state` is the right course of action, I've attached a commit doing that here. It was never implemented, so clearly no-one ever called it. Probably even if you do tackle #230, would it be by implementing this method?

cdeil:
Wait: it's actually called once from here in an if that's never executed:
https://github.com/iminuit/iminuit/blob/04fcc8af160c423f7c778b48cc41ff423925f5f7/iminuit/_libiminuit.pyx#L945-L954
If removal is appropriate, that if should be removed as well.

HDembinski:
I think we cannot remove interface that is already exposed. We can try to do that for a breaking 2.0 release.

We should implemented `get_initial_param_state`. Fortunately, that's easy because the code is already there. It is in `_prepare_initial_param`. Similarly, code in `_prepare_param` is duplicated in `get_param_states`. These functions need to be merged.

cdeil:
> I think we cannot remove interface that is already exposed.

In this case we can, because it's clear that no-one ever called it. They would have gotten a `NotImplementedError`. So in this case, you have the option to change API or put a new method, whatever you think best.

HDembinski:
Ok, let me be more clear: Yes, could in principle remove `get_initial_param_state`, but it makes more sense to put the code of `_prepare_initial_param` there and expose this for unit testing.

My statement that we cannot remove interface is related to this
>  If removal is appropriate, that [get_param_state] if should be removed as well.

HDembinski:
See #249.

HDembinski:
Closing this, as it was addressed by #249

Issue #242, opened 2018-06-20
Add test for hesse for function with gradient


cdeil:
Cython coverage (#158) shows that hesse for a function with gradient isn't covered by a test:

https://github.com/iminuit/iminuit/blob/04fcc8af160c423f7c778b48cc41ff423925f5f7/iminuit/_libiminuit.pyx#L619-L624

@HDembinski - Can you please add a test for this?

HDembinski:
Is addressed in PR, closing...

Issue #241, opened 2018-06-20
Add test for fix option in from_array_func


cdeil:
Cython coverage (#158) shows that this function isn't covered by a test:
https://github.com/iminuit/iminuit/blob/04fcc8af160c423f7c778b48cc41ff423925f5f7/iminuit/_libiminuit.pyx#L484

@HDembinski - Worth adding a test?
(if no, please add `#pragma: nocover`)

HDembinski:
Is addressed in PR, closing...

Issue #239, opened 2018-06-20
Add test for is_fixed


cdeil:
Cython coverage (#158) shows that this function isn't covered by a test:
https://github.com/iminuit/iminuit/blob/04fcc8af160c423f7c778b48cc41ff423925f5f7/iminuit/_libiminuit.pyx#L814-L822

Actually the `%s` in the `RunTimeError` message that isn't filled is a bug, no?

@HDembinski - You or me?

HDembinski:
I will add a test and fix the bug.

HDembinski:
Is addressed in PR, closing...

PR #248, opened 2018-06-21
Addresses missing test coverage

HDembinski:
We want moar tests.

Addresses #242, #241, #239.

### DIFF
--- a/iminuit/tests/test_iminuit.py
+++ b/iminuit/tests/test_iminuit.py
@@ -19,7 +19,7 @@ class Func1:
         pass
 
     def __call__(self, x, y):
-        return (x - 2.) ** 2 + (y - 5.) ** 2 + 10
+        return (x - 2.) ** 2 / 4.0 + (y - 5.) ** 2 + 10
 
 
 class Func2:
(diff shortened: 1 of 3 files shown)


HDembinski:
@cdeil In case you want to review this, the diff may look very complicated, but the new code is much simpler and more clear than the old one.

iminuit/tests/test_iminuit.py
```diff
@@ -246,63 +271,31 @@ def f(x, y):
...
-    assert_allclose(val['z'], 7)
-    assert_allclose(m.fval, 10. + 2.5)
-    free_param = m.list_of_vary_param()
-    fix_param = m.list_of_fixed_param()
-    assert 'x' in free_param
-    assert 'x' not in fix_param
-    assert 'y' in fix_param
-    assert 'y' not in free_param
-    assert 'z' not in fix_param
+    for grad in (None, func3_grad):
```
cdeil: Use https://docs.pytest.org/en/latest/parametrize.html here instead of for loop in the test?

HDembinski: Good point, I just read about this yesterday. I love pytest, I must say. It is really pythonic.

HDembinski:
I use `pytest.mark.parametrize` now to test various conditions, which is very elegant. Once the CI tests ran through, I will merge this.

PR #249, opened 2018-06-22
Issue 163 get initial param state

HDembinski:
This became another large patch, because I wanted to fully test any impact of what I did to `get_param_states`.

At the core, this patch removes `get_initial_param_state` and adds an optional keyword `initial` to `get_param_states`. If you call `get_param_states(initial=True)`, you get the initial parameter state, so exactly what `get_initial_param_state` was supposed to do.

I removed the two internal methods `_prepare_initial_param` and `_prepare_param`. The latter was identical to `get_param_states`. The former turned out to produce an inconsistent struct under certain conditions. I reimplemented the function in a consistent way.

Then I added unit tests for both versions of `get_param_states`. The method is also used by the frontends to print output on the console and in IPython, so I wanted to check that as well. When I looked into this, I found that the output of these frontends was not checked at all. They were only checked to be executable without errors. So I added new tests for the frontend output, which was a bit complicated because I had to capture and redirect output generated by these frontends. To make the frontends testable, I modified the class definition in a suitable way and added an abstract base class for the frontends, which enforces the interface.

While testing, I found a few minor cosmetic issues with the output generated by the frontends, which I fixed. The new output looks prettier and more consistent now.

Finally, I simplified the `Struct` class that iminuit uses to hold structured data such as the one provided by `get_param_states`. I also added a test for the Struct behavior.

### DIFF
--- a/iminuit/tests/test_frontend.py
+++ b/iminuit/tests/test_frontend.py
@@ -1,4 +1,11 @@
 from __future__ import (absolute_import, division, print_function)
+# Here we don't import unicode_literals, because otherwise we cannot write
+# in "\usepackage" and similar in the string literals below.
+# The problem is an inconsistency regarding the effect of
+# from __future__ import unicode_literals in Python 2 and 3, which is
+# explained here:
+# https://stackoverflow.com/questions/7602171/unicode-error-unicodeescape-codec-cant-decode-bytes-string-with-u
+# We want the same code to work in Python 2 and 3 and chose this solution.
 import sys
 from iminuit import Minuit
 import iminuit.frontends.html as html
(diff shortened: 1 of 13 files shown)


iminuit/tests/test_frontend.py
```diff
@@ -1,35 +1,560 @@
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
+from __future__ import (absolute_import, division, print_function)
```
cdeil: Suggest to add one comment line:
```
# Here we don't import unicode_literals, because it's easiest to put in "\usepackage" below
# See https://stackoverflow.com/questions/7602171/unicode-error-unicodeescape-codec-cant-decode-bytes-string-with-u
```

HDembinski: Ok, good point.

PR #250, opened 2018-06-22
Pep8

HDembinski:
pep8 everywhere based on flake8 linter


HDembinski:
I think this requires no reviewing. I will wait for the CI checks to run through and then merge this.

PR #251, opened 2018-06-27
Better tutorial

HDembinski:
I fixed, updated and extended the tutorial. Some stuff in the basic tutorial I moved into an "advanced tutorial".

I found minor issues in the frontends while doing so, which affect printing information in the console and the notebook. These I fixed as well. I also added a few new tests for the `describe` function, because I noticed an inconsistency for functors with variable arguments. This is also fixed.


cdeil:
@HDembinski - Awesome!

- Suggest to add a mention of the new `minimize` function in the first tutorial, and also add this function in the API docs: http://iminuit.readthedocs.io/en/latest/api.html
- Suggest to move the Cython example to the second tutorial. I think the fraction of users using Cython is very small, compared to the number of users that learn iminuit and have to understand how to write their likelihood and interact with Minuit. Giving a super short Cython example in the notebook is not so useful, only a longer example also teaching them how they would then move that to a `pyx` and compile and use it. But that's not for the first tutorial IMO (which is already super long).
- I think a mention / example of `merrors` in the first tutorial is missing?
- You show `least_squares_np` very early, but only show how to use it with Minuit much later. I would suggest to only short this later, where you explain the second function interface. Given how long the basic tutorial is, I think even splitting off a second tutorial "iminuit with Numpy" that explains the second / alternative way to write a function and use it would be better. People are going to choose one of the other anyways, and if they use Numpy, they can only learn the second. I still think the first / original method should remain first in the docs, even if you prefer the second.
- I see you've made changes in the frontend "Para Err" -> "Sym. Err" and "Parabolic error" to "HESSE Error" and there's a mix of "Minos Error" and "Error+" / "Error-". Suggest to just use "Error" and "Error+", "Error-" in the printouts. The other ones look annoyingly long, and I think the best way to teach this is to say "HESSE" and "MINOS" are algorithms (and thus methods you can call to execute), but what comes out is `.error` ("Error" in the printout) and `.merrors` ("Error+" and "Error-" in the printout). There could still be a docs section or link to James ~ 1970 Interpretation of errors, I just mean keep the printout short and simple and clear.
- I don't have much more time for review. Suggest to make whatever changes you fancy and merge when ready from your side. If I or others are unhappy about something, we can open another PR in the future.

PR #254, opened 2018-06-28
Fix issue 198

HDembinski:
Changes the implementation of Minuit.draw_contour to use Minuit.mncontour, which returns an exact contour, instead of Minuit.mncontour_grid, which returns an approximate contour, which is really bad if the parameters are highly correlated.

Technically, this is a breaking change, because I change the return value of draw_mncontour. I do not expect many people to actually use the return value. I could also break the arguments of Minuit.draw_mncontour, because it contains now unused keywords, but I decided against it to retain backward compatibility.


HDembinski:
Ah crap, I branched from the point. Closing this.

Issue #198, opened 2017-06-26
pixellation issues in contour plots in highly covariant variables


johnsmorgan:
I am using iminuit to do some fitting with very strong covariance between variables and I'm getting a lot of contour plots that look like this by default:
![image](https://user-images.githubusercontent.com/3793500/27530162-0a475536-5a8a-11e7-9a97-e22e6dee28cb.png)

This should show a thin ellipse. It doesn't due to the pixellation of the underlying data
![image](https://user-images.githubusercontent.com/3793500/27531702-357f5004-5a90-11e7-800b-85d6b7ecd8f3.png)

If I zoom in with nbin=40 and bound=0.5, the true covariance is clearly well-behaved.
![image](https://user-images.githubusercontent.com/3793500/27531488-5e2c4e68-5a8f-11e7-8e28-3a627142e217.png)
 
I would suggest having an odd number as the default for 'bins' as this at least ensures a central peak. 
![image](https://user-images.githubusercontent.com/3793500/27531057-ca5721be-5a8d-11e7-813d-f8b88fd119f4.png)

Having the option to add the raster image (via imshow) under the contours might also be useful option, and perhaps should be considered as the default.

cdeil:
@johnsmorgan - Apologies for not responding here.

I'm not sure what exactly you mean here or how to devise a simple test case.

Are you just suggesting to change the default for `bins=30` in
http://iminuit.readthedocs.io/en/latest/api.html#iminuit.Minuit.draw_mnprofile
to e.g. `bins=31`? Would that help in your case?

HDembinski:
I think I understand the issue. Using uneven bin number may improve the situation but does not solve the underlying problem. The way how the contour plot is made is wrong. I need to have a closer look into draw_mnprofile to fix this properly. The MINOS algorithm is able to draw exact contours. We should not use these brittle approximations.

cdeil:
@HDembinski - OK.

But there's a ton of methods to contour and parameters.

As always, my suggestion would be to mainly improve the docs, and refer them to scikit-image and MPL contouring, advertise these plot methods in iminuit only as quick-look.

HDembinski:
And as always, I don't want to be responsible for a software that gives such dangerously wrong results.

HDembinski:
Addressed in PR #255

cdeil:
Should be fixed by #255 from @HDembinski .

Issue #30, opened 2012-12-26
Ambiguous contours output


mattbellis:
When using the contours function with subtract_min=True and plotting using matplotlib's contours function, I get multiple lines that extend beyond the sigma (bound) of 1.0 that I thought I set.

PR #255, opened 2018-06-28
fix for issue 198, draw an accurate MnContour contour instead of some…

HDembinski:
Addresses #198. Changes the implementation of Minuit.draw_contour to use Minuit.mncontour, which returns an exact contour, instead of Minuit.mncontour_grid, which returns an approximate contour, which is really bad if the parameters are highly correlated.

Technically, this is a breaking change, because I change the return value of draw_mncontour. I do not expect many people to actually use the return value. I could also break the arguments of Minuit.draw_mncontour, because it contains now unused keywords, but I decided against it to retain backward compatibility.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -25,7 +25,6 @@ These are the things you will use a lot:
     Minuit.mnprofile
     Minuit.draw_mnprofile
     Minuit.mncontour
-    Minuit.mncontour_grid
     Minuit.draw_mncontour
 
 Minuit
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -1351,60 +1351,16 @@ cdef class Minuit:
 
         return xminos, yminos, cerr.Points()  #using type coersion here
 
-    def mncontour_grid(self, x, y, bins=100, nsigma=2, numpoints=20,
-                       int sigma_res=4, edges=False):
-        """Compute gridded minos contour.
-
-        **Arguments:**
-
-            - **x**, **y** parameter name
-
-            - **bins** number of bins in the grid. The boundary of the grid is
-              selected automatically by the minos error computed. Default 100.
-
-            - **nsigma** number of sigma to draw. Default 2
-
-            - **numpoints** number of points to calculate mncontour for each
-              sigma points(there are sigma_res*nsigma total)
-
-            - **sigma_res** number of sigma level to calculate MnContours
-
-            - **edges** Return bin edges instead of mid value(pass True if you
-              want to draw it using pcolormesh)
-
-        **Returns:**
-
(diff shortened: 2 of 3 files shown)


HDembinski:
After discussing with @cdeil offline a bit, I dropped backward compatibility.

I removed `Minuit.mncontour_grid` which is redundant with `Minuit.contour`. Both scan the function over a grid. I also removed the now obsolete keyword args for `Minuit.draw_mncontour` (bins and sigmares)

cdeil:
I looked at the various contour and plot contour methods and functions for 3 min, and I'm still confused what's what.

@piti118 - Could you please have a look and confirm the change here is a good one?

(@HDembinski - If we don't hear back, I'll have a closer look soon and merge)

cdeil:
> I removed Minuit.mncontour_grid which is redundant with Minuit.contour. Both scan the function over a grid.

Are you sure they do the same thing?

Or was one a "slice" through the likelihood (keeping extra parameters fixed) and the other a "profile likelihood" (re-optimising extra parameters after adjusting x / y)?

https://iminuit.readthedocs.io/en/latest/api.html#iminuit.Minuit.contour
> return contour of migrad result obtained by fixing all others parameters except x and y which are let to varied.

HDembinski:
> Are you sure they do the same thing?
Ok, when you have more than two floating parameters, they don't do the same.

Nevertheless, the removed `mncontour_grid` calls `mncontour` internally, so we still get the same behavior only now without the pixelisation.

HDembinski:
Look at what the code in `mncontour_grid` did, you can see it in the diff. It took an exact contour from Minos, then interpolated the exact contour linearly over a discrete grid. Then this grid was handed over to matplotlib again, which turned the imperfect grid representation back into an imperfect approximation to the original contour. This is madness.

HDembinski:
This PR is a clear improvement for `draw_mncontour`, I think this not controversial.

We need to decide whether we should keep `mncontour_grid`. Since it is not doing the same as `contour`. It is not redundant, but it has no statistical application. I am an expert of likelihood and least-squares fitting, and I know that the only thing that's useful are the contours, not the full 2D profile.

If `mncontour_grid` should be kept - and I am against that - we should open an issue for the next release, because `mncontour_grid` should not be implemented by interpolating contours from `mncontour`, but it should minimize the function properly for each point of the grid.

cdeil:
Looks like `mncontour_grid` was introduced in #57 by @piti118 in response to issue #30 by @mattbellis in 2012; I think this might have been a hack to work around some matplotlib contour plotting issue, but from the description it's not really clear.

@HDembinski - Your change here seems like a good one to me. Merging now.

If we get any issue reports about this plotting in the future, we can always adjust again. I don't think we have to be super-script with backward-incompatible changes to plotting code.

Issue #230, opened 2018-05-14
change current parameter state


HDembinski:
There seems to be no way to change the current parameter state in an existing Minuit instance.

Changes made to `m.fitarg[name]` or `m.values[name]` are ignored. This is extremely confusing, because these values should not be settable, if they are read-only.

Changes in `m.fitarg`, `m.values`, and `m.errors` should actually update the internal state of Minuit.

Related to #163 #105

cdeil:
@HDembinski - I'm going through the issues with no milestone, setting something.

Is this a v1.3 item? If no, please move to a later milestone.

Issue #105, opened 2013-12-03
fixing parameters


marcocamma:
Hi,
it would be great to have the possibility of fixing parameters like

``` python
m=iminuit.Minuit(f)
m.fixed["x"] = True
```

a bit like pyminuit

Thanks for your great work,
marco

cdeil:
@marcocamma @bungernut Sorry that this got lost ... I agree a better interface to fix / free parameters is needed in iminuit.

My understanding based on the [Minuit class API docs](http://iminuit.readthedocs.org/en/latest/api.html#minuit) is that it **is** possible to fix parameters via the `fix_` keywords on construction: `m = Minuit(f, fix_x=True)`. It's even possible to interact with those flags via the `fitarg` member:
```python
>>> from iminuit import Minuit
>>> def f(x, y, z):
...         return (x - 2) ** 2 + (y - 3) ** 2 + (z - 4) ** 2
... 
>>> m = Minuit(f, pedantic=False)
>>> m.fitarg
{'fix_z': False, 'y': 0.0, 'z': 0.0, 'fix_y': False, 'error_x': 1.0, 'error_z': 1.0, 'limit_x': None, 'error_y': 1.0, 'limit_y': None, 'fix_x': False, 'limit_z': None, 'x': 0.0}
>>> m.fitarg['fix_x'] = True
>>> m.fitarg['fix_x']
True
```

I didn't look yet if this actually works and if this is covered by tests.
I also don't know yet if this is related to #163 or independent.

At the very least this feature should be documented in a high-level Sphinx docs section.
Probably it would also be possible to improve the API, i.e. introduce a Parameter class like `lmfit` or `Sherpa` or `Astropy` has it and expose it as an attribute on `Minuit` that makes it easy to set / get values / errors / fix flag, instead of having to go via `fitarg` and `fix_` strings.

Let's wait for @piti118 to comment first.

I'm not sure if someone will have the time to implement this though ... help needed!

piti118:
One side limit is supported for quite a while. 

We could add method that wraps around fitarg. That should be easy enough.

marcocamma:
It does not look like the suggested method m.fitarg["fix_x"] = True is working as shown by the code below:

```python
from __future__ import print_function
import iminuit
import numpy as np

print("Iminuit version",iminuit.__version__)

def generateData(N=100,order=2,err=0.1):
  x = np.arange(N)
  coefs = np.random.random(order+1)
[...]

PR #256, opened 2018-06-29
Modify parameter state

HDembinski:
Addresses #230 #105 

It is one of the key features of Minuit is that you can modify the parameter state. A common approach is to fix some parameters initially, then minimize only free parameters, then release all parameters and fit all together. Example:

```
m = Minuit(some_func, x=1, y=2, fix_y=True)
m.migrad()  # fit only x
m.fixed['y'] = False
m.migrad()  # resume fit, now with floating y
```

So far this was not possible, because there was no way to interact with the parameter state in Minuit. The proposed workaround was to create a new `Minuit` instance. This was not intuitive for many people, and actually a step back from precursors like pyminuit, which allowed to modify the fit state.

So I fixed this by changing the classes that are returned when you call `Minuit.args`, `Minuit.values`, `Minuit.errors` from tuples and dicts to tuple-like and dict-like objects which are actually views of the internal parameter state. And I added a new field `Minuit.fixed`, a dict-like object to fix or release parameters. Now the code example above works as expected and many new ways of interactively working with Minuit become more straight-forward.

So these new views mimic a normal dict by their interface, but when you set something it actually changes the internal cpp state in the `Minuit` instance. Using views actually makes the code in `Minuit` simpler in several places, because it is not necessary to manually update `args`, `values`, `errors`, and `fixed` to the internal cpp state. They are automatically kept consistent, which removes the burden from the programmer to make sure they are.

If you want to make real copies of `args`, `values`, etc. you can do that. There is a `.copy` method, and you can also do something like `dict(m.values)` to get a copy.

The new classes have docstrings and a new test was written to check the new behavior. The change is fully backward-compatible.

### DIFF
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -782,7 +782,6 @@ cdef class Minuit:
         """
 
         cdef MnHesse*hesse = NULL
-        cdef MnUserParameterState upst
         if self.print_level > 0: self.frontend.print_banner('HESSE')
         if self.cfmin is NULL:
             raise RuntimeError('Run migrad first')
@@ -799,7 +798,7 @@ cdef class Minuit:
                 self.last_upst,
                 maxcall
             )
-        if not upst.HasCovariance():
+        if not self.last_upst.HasCovariance():
             warn("HESSE Failed. Covariance and GlobalCC will not be available",
                  HesseFailedWarning)
         self.refreshInternalState()
@@ -997,7 +996,10 @@ cdef class Minuit:
         return self.np_matrix(correlation=False, skip_fixed=False)
 
     def is_fixed(self, vname):
-        """Check if variable *vname* is (initially) fixed"""
+        """Check if variable *vname* is fixed.
+
+        Note that `Minuit.fixed` was added to fix and release parameters.
+        """
         return self.fixed[vname]
 
     #dealing with frontend conversion
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -100,61 +100,38 @@ cdef check_extra_args(parameters, kwd):
 
 
 # Helper classes
-cdef class NameIter:
-    """Iterator over parameter names in correct order"""
-    cdef MnUserParameterStatePtr _state
-    cdef unsigned int i
-    cdef unsigned int n
-
-    def __init__(self, npar):
-        self.i = 0
-        self.n = npar
-
-    def __iter__(self):
-        return self
-
-    def __next__(self):
-        if self.i == self.n:
-            raise StopIteration
-        cdef const char* s = self._state.Name(self.i)
-        self.i += 1
-        return s
-
-
 cdef class BasicView:
     """Dict-like view of parameter state.
 
     Derived classes need to implement methods _set and _get to access
     specific properties of the parameter state."""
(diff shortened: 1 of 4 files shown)


cdeil:
From the description, this sounds great to me. I plan to have a look at the diff and try this out today.
Could you please rebase this branch and resolve the git conflict?

iminuit/_libiminuit.pyx
```diff
@@ -24,10 +24,207 @@ np.import_array()
...
+            #need to set value again
+            #taking care of internal/external transformation
+            state.SetValue(i, val)
+            state.SetError(i, err)
+
+        if fitarg['fix_' + pname]:
+            state.Fix(i)
+
+
+cdef auto_frontend():
```
cdeil: A question about this `auto_frontend`: why do we set HTML frontend like this? Won't this be set if we're in `ipython` terminal where no HTML output should be used? Shouldn't the check be rather for Jupyter somehow? (If needed at all, I think they have a special method based interface instead to emit rich repr content, maybe we should use that?

It's not a change you made here, just a question because I see this code now appear in the diff.

cdeil:
When can / should one modify the state?
Only between `migrad`, `hesse`, `minos` calls?
Or also from the call back function, during the execution of `migrad` or `hesse` or `minos`?

There's no high-level docs change here; probably this common usage pattern should be mentioned with one example, no?
And a test added that state changes from the callback function at least don't lead to a crash or segfault or very hard to understand error message?

cdeil:
@HDembinski - As you know, I'm not familiar enough or have time to review the actual implementation. These proxy objects to sync the state look OK at first glance, but they are non-trivial and I guess it could be possible for bugs to hide in some cases. So just as a general suggestion, you could add a few more tests that exercise the state changes, e.g. also from the call back funciton, or in an example that includes hesse or minos, or generally to increase test coverage.

Looking at this, I wonder if there's a way to change the parameter state related to the limits. Is it possible to set something about the limits, or are those read-only.

The `is_const` is just a duplicate of `is_fixed` that we have to keep for backward-compatibility? Or is it something that's actually of interest?
```
In [9]: m.get_param_states()
Out[9]: 
[{'number': 0,
  'name': 'x',
  'value': 2.0000000000047327,
  'error': 1.0000000000000409,
  'is_const': False,
  'is_fixed': False,
  'has_limits': False,
  'has_lower_limit': False,
  'has_upper_limit': False,
  'lower_limit': None,
  'upper_limit': None},
 {'number': 1,
  'name': 'y',
  'value': 3.000000000007099,
  'error': 1.0000000000002762,
  'is_const': False,
  'is_fixed': False,
  'has_limits': False,
  'has_lower_limit': False,
  'has_upper_limit': False,
  'lower_limit': None,
  'upper_limit': None},
 {'number': 2,
  'name': 'z',
  'value': 4.000000000009465,
  'error': 0.9999999999998322,
  'is_const': False,
  'is_fixed': False,
  'has_limits': False,
  'has_lower_limit': False,
  'has_upper_limit': False,
  'lower_limit': None,
  'upper_limit': None}]
```

iminuit/_libiminuit.pyx
HDembinski: Not a change I made, but it is something I would like to see fixed as well. Opened an issue about it.

PR #245, opened 2018-07-04
Test notebooks in CI

cdeil:
If we keep notebooks, I think we should test them in CI.
Otherwise it's very hard to maintain the content, even for a stable package like this one.

See #15. I'll keep that closed, because that was a very old attempt to implement that. By now the Jupyter tooling is better. Assigning this to myself under the v1.3 milestone.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
       pip uninstall -y cython;
     fi
   - if [[ $BUILD != NO_EXTRA_DEPS ]]; then pip install IPython numpy matplotlib; fi
-  - if [[ $BUILD == DOCS ]]; then pip install matplotlib sphinx sphinx_rtd_theme, jupyter; fi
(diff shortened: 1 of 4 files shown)


cdeil:
@HDembinski - Can you please try out and review this branch?

For me locally, the Cython cell in the hard core tutorial doesn't work when executed like this:
https://gist.github.com/cdeil/bc25b76017361944f4cb9da41a50911d#file-gistfile1-txt-L121

Not sure how to make it better.

cdeil:
I see the same error from the hardcore tutorial when executing it via `jupyter lab`. I think it's just broken? @HDembinski - Does it work for you?

I'd prefer to handle that in a separate issue from adding CI for notebooks, skipping that notebook test for now: 14f583d

cdeil:
Green light from CI. Merging now.

PR #267, opened 2018-07-06
Add CI that builds and tests sdist

cdeil:
@HDembinski - here's the local fails I got when running from the `sdist` for `test_console` and `test_cython_embedsig`: https://gist.github.com/cdeil/ea3bbea7f5a72b474e1bbba27c24ec42

This should be understood / fixed before making a release.

The first step should be to add a CI test that builds and tests the sdist on travis-ci and appveyor, with Python 2 and 3. That's the main thing that needs to work, not just from the dev folder. Maybe some current builds can be removed.

I see this one:
https://github.com/iminuit/iminuit/blob/3f98a13a2caf908f21c087c4c29044921eafa193/.travis.yml#L25-L26
but it's not even running the tests currently:
https://travis-ci.org/iminuit/iminuit/jobs/400390456#L1043

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,9 @@ matrix:
         - python: 2.7
           env: BUILD=SDIST
 
-        - python: 3.6
-          env: BUILD=CONDA
+          # TODO: add this (travis-ci not running today, so disabling for now)
+#        - python: 3.6
+#          env: BUILD=CONDA
 
 # TODO: add osx builds; (setting up multi-os envs isn't trivial, see below)
 #        - os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,10 @@ sudo: false
 matrix:
     include:
         - python: 2.7
-          env: BUILD=TEST
-
-        - python: 2.7
-          env: BUILD=DOCS
+          env: BUILD=ALL
 
         - python: 3.5
-          env: BUILD=TEST
-
-        - python: 3.5
-          env: BUILD=DOCS
+          env: BUILD=ALL
 
         - python: 3.5
           env: BUILD=COVERAGE
@@ -25,7 +19,13 @@ matrix:
         - python: 3.5
           env: BUILD=NO_CYTHON
 
-        # TODO: add tests on Mac
+        - os: osx
+          python: 2.7
+          env: BUILD=ALL
+
+        - os: osx
+          python: 3.5
+          env: BUILD=ALL
 
         # TODO: add tests for Python from conda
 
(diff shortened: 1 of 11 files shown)


cdeil:
@HDembinski - I've started to edit `.travis.yml` in 3b8497b, but I think it'll take quite some trial and error to get it to work. I'll try to do it this afternoon.

HDembinski:
Yes debugging Travis is a pain

HDembinski:
I installed docker on my maschine, which allows you to download an run the travis VM on your computer. That helps a little. It is still a cumbersome process, the steps can be found by googling.

HDembinski:
After some hours of reading and debugging I can say that this is a wontfix for now.

The errors come from the fact that any data files such as cyfunc.pyx are NOT installed when the package is installed. The files end up correctly in the archive, but are not installed in your local package folder where the .py and .so files end up. 

This is correct according to the documentation of setuptools. One needs to set `include_package_data=True` in the `setup(...)` call to not only put the files in MANIFEST.in in the archive, but to also install them. See:
https://stackoverflow.com/questions/13307408/python-packaging-data-files-are-put-properly-in-tar-gz-file-but-are-not-install

Now, this flag does not solve our problem. Once `include_package_data=True` is set, installing iminuit fails when installing with this error message
``
    error: Error: setup script specifies an absolute path:
    
        /tmp/pip-req-build-o3Rd34/iminuit/Minuit/src/AnalyticalGradientCalculator.cxx
    
    setup() arguments must *always* be /-separated paths relative to the
    setup.py directory, *never* absolute paths.
``

Of course there is no such thing as an absolute path set anywhere in our configuration. This is a bug in setuptools, which was also found by others:
https://stackoverflow.com/questions/18085571/pip-install-error-setup-script-specifies-an-absolute-path

I will modify the tests for now so that they run without cyfunc.pyx and friends. In the longterm, we should try to find a why to include data files in the package.

HDembinski:
Your error in `test_console` revealed a minor bug in the test. I will fix that one in a separate PR.

HDembinski:
@cdeil I created PR #269 with the fixes for these failures. I would be cool to merge this PR before we tag the release and upload to PyPI.

cdeil:
I was also trying to add a conda build here, because that does a lot of new checks.
But travis-ci stopped running, it's not starting builds.

However, the last build that ran did successfully make and install and run tests from the sdist.

So I'm commenting out the conda build in `.travis.yml` and merging this PR now.

@HDembinski - If you want to make an iminuit release, please just go ahead and let me know if you have any questions. Otherwise, talk to you on Monday.

PR #281, opened 2018-07-16
Better readme

HDembinski:
Get badges back, now it works for real

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ Branch  Linux (Py-2.7, Py-3.6)  Windows (Py-2.7, Py-3.6)
 master  |travis|                |appveyor|
 ======  ======================  ========================
 
-.. |travis| image:: https://ci.appveyor.com/api/projects/status/g6vymxvu9ax34e7l/branch/master?svg=true
+.. |travis| image:: https://travis-ci.org/iminuit/iminuit.svg?branch=master
 .. |appveyor| image:: https://ci.appveyor.com/api/projects/status/g6vymxvu9ax34e7l/branch/master?svg=true
 
 iminuit works with Python-2.7 to 3.5 on Windows, Mac (currently manually tested), and Linux.
--- a/README.rst
+++ b/README.rst
@@ -1,13 +1,18 @@
 iminuit
 =======
 
-MINUIT from Python - Fitting like a boss
+**Python interface to the MINUIT2 C++ package - Fitting like a boss!**
 
-`iminuit` is a Python interface to the `MINUIT` C++ package.
+======  ======================  ========================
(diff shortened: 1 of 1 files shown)


cdeil:
Did you see status.rst? I removed the badges from the Readme, because that is used for the PyPI front page, and there I don’T think it belongs, status is only for the dev version.

HDembinski:
So what do we do then? I think we need badges on github.

HDembinski:
The readme is the first entry point for people on github. There we want to show that this project is not dead but working well and badges are the way to do it.

I value badges in the readme more than having one file that is consistent between PyPI and github.

HDembinski:
We could also do something smart and remove the badges automatically when the readme is parsed by setup.py. Perhaps that's the way to go.

cdeil:
Actually you and I are probably the only ones that care about these badges.
Note that most users will use and care about the released version, so if they for some reason (i.e. Google) find the Github page and see a successful or broken build status for master, it's not really helpful.

My suggestion would be to put it as it was, i.e. have the badges in status.rst.

If you really want them in the README for some reason, you have to write the `long_description` on the `setup` call (what PyPI shows) somehow differently, e.g. put it as a multi-line string in the setup.py.

cdeil:
@HDembinski - How should we resolve this discussion / PR?

I've had projects with badges in README and shown on PyPI before, and it was always a mess because the badges showed the CI status for the dev version, sometimes as broken, but were displayed on PyPI for a stable version that was perfectly fine.

What we currently have is nice, no? https://pypi.org/project/iminuit/

Suggest to keep as-is, or add one line to the README for developers: "For the continuous integration status of iminuit, see the `status.rst` file on Github." if you think that's useful.

Changing the way we set `long_description` and making it different from the README is extra effort / complexity for little gain, no?

HDembinski:
Let's close the request. I was thinking of writing some code that parses the README.rst in setup.py and removes the badges tables, but I don't have time to work on it and I can also make a new PR when I ever feel that this is really important.

cdeil:
> I was thinking of writing some code that parses the README.rst in setup.py and removes the badges tables

IMO setup.py shouldn't contain such code. Even if unlikely, it has the potential to break things e.g. if someone doesn't know about it and edits the README in a few years and then just sees on https://pypi.org/project/iminuit/ that the output is garbled after making a release.

HDembinski:
And I don't like that we don't have badges on github anymore, which I consider to be very useful. What other solution would you propose to get badges on github, while keeping the readme constistent between PyPI and github?

cdeil:
The badges are visible on github, you just have to do one click: https://github.com/iminuit/iminuit/blob/master/status.rst

IMO one click is acceptable for the very few people that very rarely need to look at this.

PR #283, opened 2018-08-03
fix for segfault when PythonCaller throws

HDembinski:
with tests to make sure it does not go by unnoticed anymore


cdeil:
🙈

HDembinski:
I will prepare a minor release, as this is a breaker...

PR #284, opened 2018-08-09
fix broken matrix implementation

HDembinski:
The implementation of matrix() in 1.3 was completely wrong when parameters were fixed. It was reading uninitialized memory. I am not quite sure why the tests did not catch this, as this is tested. Perhaps the test was too simple to show these issues.

### DIFF
--- a/iminuit/_libiminuit.pyx
+++ b/iminuit/_libiminuit.pyx
@@ -883,25 +883,33 @@ cdef class Minuit:
         cdef MnUserCovariance mncov = self.last_upst.Covariance()
         cdef vector[MinuitParameter] mp = self.last_upst.MinuitParameters()
 
+        # When some parameters are fixed, mncov is a sub-matrix. If skip-fixed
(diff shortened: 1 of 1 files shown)


cdeil:
Would be great to have a regression test for this, no?

Was this the ZeroDivisionError we saw with Gammapy?
Did you ever manage to trigger it with a simple test case?

HDembinski:
Yes, would be nice to have a regression test. I don't know how to write one, because the result you got is not 100 % reproducable. The previous implementation was reading random memory. Some of that memory was interpreted as the diagonal elements of the covariance matrix. When you compute the correlation, you divide by the diagonal elements which are zero. Hence the DivideByZero.

Once this is fixed, it is not reproducable anymore.

HDembinski:
You can also be lucky and read some memory that is not zero, and then you don't get a DivideByZero (although the covariance matrix is still nonsense).

PR #297, opened 2018-08-29
Python error report

HDembinski:
When user functions raises an exception, report that original exception in addition to the extra information added by the C++ code.


cdeil:
Thanks! Regression test?

HDembinski:
Stupid Python3

Issue #294, opened 2018-08-29
Show exception from likelihood function?


cdeil:
@robertazanin and I were having issues debugging why the Python kernel was crashing when running MIGRAD via Gammapy from Jupyter. The problem was that a Python exception occured in the FCN and she was still using iminuit 1.2 .

Still, with iminuit 1.3 I still don't see what exception occurred.

Using this minimal example
```
from iminuit import Minuit

def f(x, y, z):
    1/0

m = Minuit(f, pedantic=False)

m.migrad()
```
I get this:
```
>>> m.migrad()
**************************************************
*                     MIGRAD                     *
**************************************************

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iminuit/_libiminuit.pyx", line 757, in iminuit._libiminuit.Minuit.migrad (/private/var/folders/t_/_mywtcj146lbk2c99bnxw7z40000gp/T/pip-install-vbxbryr0/iminuit/iminuit/_libiminuit.cpp:9736)
RuntimeError: Exception Occured
fcn is called with following arguments:
    x = +0.000000
    y = +0.000000
    z = +0.000000
```

@HDembinski - Is it possible to either raise, or at least print the exception traceback, to help users debug what is going on?

HDembinski:
Addressed in #297

cdeil:
I still don't see a ZeroDivisionError or a NameError appear in iminuit master, after #297 .

There's also no regression test to show that it works. I'm happy to add that using [pytest.raises](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions), but I have no idea how to fix the exception handling.

```
>>> from iminuit import Minuit
>>> 
>>> def f(x, y, z):
...     1/0
... 
>>> m = Minuit(f, pedantic=False)
>>> 
>>> m.migrad()
**************************************************
*                     MIGRAD                     *
**************************************************

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iminuit/_libiminuit.pyx", line 775, in iminuit._libiminuit.Minuit.migrad
    self.cfmin = call_mnapplication_wrapper(
RuntimeError: exception was raised in user function
fcn is called with following arguments:
    x = +0.000000
    y = +0.000000
    z = +0.000000

>>> def f(x):
...     try:
...         1/0
...     except:
...         spam
... 
>>> m = Minuit(f, pedantic=False)
>>> m.migrad()
**************************************************
*                     MIGRAD                     *
[...]

PR #298, opened 2018-08-31
fixing exception report for python 2 and python 3

HDembinski:
This is now really fixed and works in Python-2.7 and Python-3.7. Tests check that the original error message is included.


HDembinski:
This should fix #294

cdeil:
@HDembinski - I'm off now. Please merge if CI passes and I'll try in the next days.

PR #300, opened 2018-09-06
WIP: Adding standalone minuit2

henryiii:
This adds the standalone version of current Minuit2 source.

#### Changes:
* Requires C++11 because Minuit2 does
* Added guide for getting Minuit2 source
* Dropped and simplified some bits of the setup.py - based on PyBind11's setup.py

#### Problem(s):
Currently this fails to build because Cython wants `ROOT::Minuit2::ContoursError` to be default constructible, which it is not. Someone with more recent Cython experience can probably fix this faster than I can.

Until this is fixed this is little more than a proof of principle (if that).


HDembinski:
Dear Henry, thank you very much. We discussed about this already in private and I think it is good when iminuit uses standalone Minuit2. But I would like to include it as a submodule then and not as a source copy. Otherwise I don't see much of an advantage.

HDembinski:
More questions: the original Minuit2 code is not C++11, how did C++11 creep in? I am personally all for supporting C++11, but I vaguely remember that it gave us some trouble so I avoided it.

henryiii:
It would be simple to include it as a submodule from GooFit/Minuit2 (ROOT is not interested in providing an official git submodule). The problems with the build would still need to be fixed, however.

Putting that back in is fine, but I wanted to simplify things at first as much as possible to find the bugs (currently the one I mentioned).

HDembinski:
Ok, so we would try to get this PR running and if that works, we replace the source copy with a git submodule link.

henryiii:
ROOT is C++11, so I'm guessing that the components of ROOT have been updated in the last few years. It is possible (since I'm building the entire Miniut2, whereas the old version seems to be a very stripped down and modified version) that I'm simply building too much, though.

henryiii:
> and if that works, we replace the source copy with a git submodule link

Yes, exactly.

HDembinski:
I opened an issue for this PR #301

HDembinski:
> Yes, exactly. And add back the time-saver recompile part.

Everything should be fixed already in this PR, so that once we merge this PR, we only need to replace the source code copy of Minuit2 with a git submodule link.

cdeil:
@henryiii - Thank you very much!

> This adds the standalone version of current Minuit2 source.

How is it different from the version we currently have, and why is it better?

I thought @HDembinski you updated the vendored version of Minuit2 in iminuit to the latest from ROOT a few months ago already.

@henryiii or @HDembinski - I guess it's clear to you, but can you please add a few lines on the iminuit docs page about which version of Minuit2 we are using?
https://iminuit.readthedocs.io/en/latest/about.html

Requiring C++11 now is a change, but I guess shouldn't be a problem.

I would suggest to vendor a copy instead of introducing a git submodule, keep it simple and independent But also on this point: works either way, @HDembinski - if you want a git submodule pointing to GooFit or somewhere else, OK.

henryiii:
> I would suggest to vendor a copy instead of introducing a git submodule

We can use [git subtrees](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree), if someone wants to read up on how to do it (I prefer submodules).

> I thought @HDembinski you updated the vendored version of Minuit2 in iminuit to the latest from ROOT a few months ago already.

This is, as far as I can tell, a modified and stripped out version of Minuit2. Most of files have changes in them relative to the source code for Minuit2. (I could be wrong). The version here is exactly the same as ROOT (it's the latest master at the moment).

> Everything should be fixed already in this PR

Everything should be fixed by the time the WIP is removed, anyway - this is not intended for merging until then.

> we only need to replace the source code copy of Minuit2 with a git submodule link.

Yes, the GooFit/Miniut2 submodule is kept up by the exact same procedure used to add these files (though I left a couple of files out in this copy, like the tests).

HDembinski:
@cdeil I think the version that Henry put in is effectively the same version that we had. The diff is so large, because the directory structure is different, I think. So, nothing changes for now, but in the long run we profit from this move, as it simplifies our life.

I just added the ROOT version to the About page. It was listed previously in the Changelog.

We cannot directly make a git submodule link to CERNs ROOT, because they don't provide a separate repo for Minuit2, but Henry has one in GooFit that he syncs with CERN's version.

cdeil:
OK. Thanks @henryiii and @HDembinski for all your work!

(I'll disable notifications for the `iminuit` repo now, I don't have time to contribute at the moment anyways.)

HDembinski:
Any chance that we can finish this PR? Otherwise I am inclined to close it.

henryiii:
I was just talking to Lorentzo about this a couple of days ago at CHEP, actually! Now that we build wheels, the C++11 in the newer Minuit2 shouldn't be a problem. I can try to update this and see what is needed in a few days. Ping me at the end of next week if I don't update this, that means I forgot.

henryiii:
(If there's an issue that requires modification to the Minuit2 source, let's try it out, then try to upstream it if possible. I don't remember the details of the failure ATM)

Minuit/README.md
```diff
@@ -0,0 +1,15 @@
+This was synced with the Minuit2 source from ROOT using the standalone build.
```
HDembinski: There should be a script to automate this, like the deleted `import_minuit.py`.

HDembinski: The information should be placed in the iminuit docs folder, to avoid accidental loss.

HDembinski:
Looking more closely at this PR. The MINUIT2 code should be included as a git submodule, not as a copy. Just copying from another source is not an improvement over the current situation.

HDembinski:
I am closing this, as it has to be redone from scratch anyway.

henryiii:
Actually, I have an idea. We can include ROOT as a submodule; the standalone iminuit2 was designed to be used inplace. We can just copy the single portion of the ROOT tree to the sdist. So:

* Git submodule all of ROOT, so any git development/installs will get ROOT. ROOT's source is not unmanageably large, and shallow checkouts can speed it up even more.
* SDists do not include more than they need from ROOT (`math/minuit2`) using a MANIFEST.in rule.

No copies or manual syncing.

Thoughts?

henryiii:
> I am closing this, as it has to be redone from scratch anyway.

That's okay, yes, it's not an incremental change.

henryiii:
I'm also trying to get Lorenzo to update the old Minuit2 page, and maybe that will have a tar.gz, would that help? That was another thing we discussed.

HDembinski:
I have successfully switched to GooFits Minuit2 in the develop branch. The problem you were facing in this PR was actually made by us, we patched on of the classes to make the Cython code work. I wrote a workaround for this, now iminuit works with a pristine Minuit2 and the code is more efficient, too, it avoids two allocations from the heap.

I could also include ROOT as a submodule, true. It is up to you, but if you keep updating GooFits Minuit2, I see no reason to switch to ROOT?

HDembinski:
I really want to have Minuit2 as a submodule, it makes living at head so much easier.

henryiii:
Okay, sounds good. Would you mind if I gave you full permissions into that module too? I don't update it as often as I should, since a new maintainer is being trained to work on GooFit. I'll do an update in the next few days and write the instructions down on an attached wiki page (since the module contents come directly from ROOT).

Agree about the submodules being much easier to update/live at head.

henryiii:
(Once I write the guide, you can copy it into the iMinuit docs folder too)

HDembinski:
Sure, thank you.

PR #314, opened 2018-09-12
Feature/contributing update

HDembinski:
I want to switch to Gitflow. The new contributing guidelines are described here.

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -32,7 +32,7 @@ To hack on `iminuit`, start by cloning the repository from `Github`_:
     $ cd iminuit
     $ git checkout develop # switch to develop branch for making changes
 
-We use `git-flow <https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow>`_. This means that all changes are applied to the `develop`, while the `master` branch only tracks releases.
+We use the popular `Gitflow workflow <https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow>`_. This means that all changes are applied to the `develop` branch, while the `master` branch only tracks releases.
 
 conda
 +++++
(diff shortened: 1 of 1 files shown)


HDembinski:
@piti118 @cdeil I am using the popular Gitflow workflow for boost.histogram and had a good experience with it, so I also wanted to propose it for iminuit, but now I got doubts, because it makes PRs slightly more complicated. You have to make sure to make PRs from the develop instead of the master. Maybe it is not worth it?

I thought this would make maintaining releases easier, but since we don't make breaking changes, we don't really have to maintain releases. And then the gitflow workflow does not really adds something.

cdeil:
I'm +10 to keep the simple workflow (basically https://trunkbaseddevelopment.com/ ) we (and most projects on Github) have used so far. I don't see a reason to introduce something more complex.

HDembinski:
@cdeil Do you have hard numbers to back up "most projects on Github"? All the Boost repos use Gitflow.

HDembinski:
But anyway, no Gitflow is ok. I will still use this PR to update the contribute.rst page. Further commits will be coming.

cdeil:
@HDembinski - Thanks for sticking with the simple master-based development workflow here, even if you prefer the master/devel git flow.

I'm not sure if there's value in introducing stable branches and bugfix backport / patch releases for iminuit. It creates significant extra work for you, and I think given that iminuit is stable and we're only fixing bugs and adding features in a backward-compat way, it's just not needed, users can always update to the latest release.

If nevertheless you want to start this for some reason, I still no need for a devel branch. Many projects (certainly Astropy, but I think Numpy, Scipy, scikit-image, scikit-learn, ...) develop on master and then backport to the stable branches. Astropy probably has the most extensive docs for their workflow, see e.g. http://docs.astropy.org/en/stable/development/releasing.html#maintaining-bug-fix-releases

Issue #337, opened 2019-05-17
Latest bugfix release contains breaking changes


mayou36:
It seems like 1.3.5 contains breaking changes, which are also not documented in the release notes

m.values does not return a dict anymore but a tuple (so e.g. the example in the README is wrong now). This breaks existing code relying on the fact that it is a mapping and not a tuple.

I would suggest to remove this release and make a minor (1.4.0) release? Or restore the mapping behavior (which seems more convenient to me).

HDembinski:
I cannot reproduce this. `m.values` does not return a dict, that's true, but this change was already made a few versions prior. The object that `m.values` returns acts like a dict, however, you should be able to use it like at dict. In addition you can use it like a tuple. By that I mean you can pass either a string of the parameter name or the parameter index.

HDembinski:
Could you send me a minimal code snipped which demonstrates the wrong behavior?

cdeil:
Something changed. When I update `iminuit` from 1.3.3 to 1.3.5, then I get these fails in Gammapy:
https://gist.github.com/cdeil/877aa2895cc26f04dce4c1c6cc40d67f

HDembinski:
So the return value of the minos call changed. That was not intended. I will investigate and fix this get the previous behavior. Thank you for the quick feedback and sorry for the inconvenience. If you update that fast, I could do a release candidate next for you to check.

mayou36:
Yes, sorry, it's not in `m.values`, but in the minimization result. Here a snippet:
```python
result = m.migrad(lambda x: x**2)
param_result = result[1]
value = param_result[0]["value"]  # this used to work
```

HDembinski:
Thank you for the minimal example, I should be able to fix this quickly.

mayou36:
Perfect, thanks a lot!

cdeil:
In Gammapy, it's failing because `minuit.get_param_states()[0]["has_limits"]` no longer works:
https://gist.github.com/cdeil/e4b9a40ea840149ce5be7325358bcd34

@HDembinski - Should I send a PR adding a travis-ci build with Gammapy now?
We don't use much from the `iminuit` API, and really a separate repo on integration testing might be better (like e.g. what Numba does: https://github.com/numba/numba-integration-testing), but probably just adding a simple build to the exiting CI is an improvement for now?

HDembinski:
[...]

PR #338, opened 2019-05-17
Add CI test with Gammapy

cdeil:
This PR adds a CI test for Gammapy.

See https://github.com/scikit-hep/iminuit/issues/337#issuecomment-493412424

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,5 +54,16 @@ matrix:
       script:
         source ci-scripts/travis_sdist_test.sh
 
+    - name: "integration: gammapy"
+      language: sh
+      install:
+        - pip install --user --upgrade cython numpy pytest matplotlib scipy
+        - pip install --user .
+        - pip install --user gammapy
+      script:
+        - cd /tmp
+        - gammapy info
+        - python -c "import sys, gammapy; sys.exit(gammapy.test())"
+
   allow_failures:
     - name: "pypy3.5: test"


cdeil:
I don't think it matters much if we test the latest stable or the Gammapy dev version. In Gammapy we have bi-monthly releases, and the goal should be for iminuit to be backwards-compat, i.e. work with all the recent releases of Gammapy from the past months.

So clearly testing more / other Gammapy versions is possible, but overall it shouldn't matter much.

@HDembinski - What are the other "big" customers - should we add any other projects that use iminuit in a similar way?

cdeil:
The test now runs and shows the issue in #337 , see
https://travis-ci.org/scikit-hep/iminuit/jobs/533805080#L944

The build still passes. I'll try putting `;` now as suggested here:
https://github.com/nipy/dipy/blob/8379bbde5bfe07ed2d6fa48f4f08953b298f62f5/.travis.yml#L6

cdeil:
The problem was that the exit status was always zero.

Now I changed to this, which should return non-zero exit status if there's a test fail:
```
- python -c "import sys, gammapy; sys.exit(gammapy.test())"
```

@HDembinski - Assigning this PR to you.
You could merge it now and have the fail in master, or first fix the regression, as you prefer.

cdeil:
It's working: https://travis-ci.org/scikit-hep/iminuit/jobs/533816072#L1193
```
The command "python -c "import sys, gammapy; sys.exit(gammapy.test())"" exited with 1.
```

henryiii:
This is great +1

HDembinski:
@cdeil Thank you! The other "consumer" is probfit, I will add an integration test later. I checked probfit before the broken release and it didn't reveal the error.

HDembinski:
@cdeil I believe that "integration tests" are an intermediate solution and not a good idea in general. They reveal deficiencies in the iminuit internal tests. Every library should test its public interface thoroughly. If that's the case, then there is no way to introduce a breaking change without breaking a test. A library shouldn't rely on its consumers for that.

henryiii:
@HDembinski I agree. Could I request that when you add an integration test for probfit, you use git rather than pypi when you pip install it? I'd like to avoid having our own CI artificially adding to our PyPI download stats (though at some point this should be fixed, pip 19.1 is supposed to report when it runs in CI).

henryiii:
PS: Not a big deal if you prefer the PyPI package, there already are other packages in Scikit-HEP that do not do this correctly either; and it's harder to do if you have a dependency of a dependency.

HDembinski:
I decided against adding integration tests to the CI. I will add some integration test scripts in the CI script folder that can be run manually before making a release, but I think it is a waste of resources to run gammapy and probfit tests regularly in CI.

HDembinski:
And when failures are discovered by doing an integration tests that iminuits own tests didn't reveal, then iminuit's tests must be extended. I did that for the API that broke gammapy.

HDembinski:
> Could I request that when you add an integration test for probfit, you use git rather than pypi when you pip install it?

Ok, but for an integration test, it is nice to install from PyPI, because you automatically get the latest release. When installing from github, you have to set the release tag manually and update it.

henryiii:
PyPI is not bad for one that doesn't run very often - and hopefully the change in pip 19.1 will help in the future. I would set it up to only run on master - you don't need to run integration tests on develop. It really should run on Travis or Azure, though, and not only on a personal computer somewhere. It's trivial to set up a new Azure pipeline for this; you can even set a date based trigger, but just adding a "skip if not on master" to travis for this line would be fine. You can add a note in CONTRIBUTING.md or somewhere that any failing integration tests need a new unit test as well to fix.

HDembinski:
I am only run them manually on a release now, so they won't be running very often.

HDembinski:
I don't think CONTRIBUTING.md is the right place, since this is information for the maintainer, not for the contributor. I will put the information somewhere, though.

PR #340, opened 2019-05-18
Adding azure

### LINKED ISSUE #334: Distribute iminuit as a wheel on PyPI
It would be quite nice if `iminuit` would distribute wheels for the library on PyPI. [Wheels](https://wheel.readthedocs.io/en/latest/) make installation much faster and given that not everyone wants to use a very bulky system like Conda (which is [currently suffering terrible speed issues](https://www.anaconda.com/understanding-and-improving-condas-performance/) that [mamba](https://github.com/QuantStack/mamba) may or may not fix) it is a great resource for many users to have binary dists available.

At the moment it appears that there are issues with the way that iminuit's CI is setup that the building of wheels is failing. For example, see the following from a [Python 3.6 job on Travis CI](https://travis-ci.org/scikit-hep/iminuit/jobs/528537856#L736-L784)

```
  x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -DWARNINGMSG=1 -I/tmp/pip-req-build-IViZEC/Minuit/inc -I/usr/include/python2.7 -c /tmp/pip-req-build-IViZEC/iminuit/_libiminuit.cpp -o build/temp.linux-x86_64-2.7/tmp/pip-req-build-IViZEC/iminuit/_libiminuit.o
  cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
  In file included from /tmp/pip-req-build-IViZEC/iminuit/PythonFCN.h:9:0,
                   from /tmp/pip-req-build-IViZEC/iminuit/_libiminuit.cpp:723:
  /tmp/pip-req-build-IViZEC/iminuit/PythonCaller.h:10:31: fatal error: numpy/arrayobject.h: No such file or directory
   #include "numpy/arrayobject.h"
                                 ^
  compilation terminated.
  error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for iminuit
```

However, if these issues can be fixed then the good news is that having the CI build and distribute wheels is [incredibly simple](https://github.com/diana-hep/pyhf/blob/b6edffae1a1cb7fc7544052506ad790a81960171/.travis.yml#L145).

If this would be considered in the future I would be willing to give time, as available, to help out on this.

henryiii:
This adds an Azure pipeline build, producing wheels and an sdist (with latest Cython). I'll need to set up the release pipeline as well to send these to PyPI, but that's entirely external. It will also be a manual step for now, rather than happening automatically on release tags. See the latest build here: https://dev.azure.com/scikit-hep/iMinuit/_build/results?buildId=786 . Closes #334 .

Minor cleanup on Travis (Python 3.7 was released, for example). The GammaPy test won't pass yet, obviously. PyTest integration and testing was slightly adjusted for consistency.

### DIFF
--- a/.ci/azure-steps.yml
+++ b/.ci/azure-steps.yml
@@ -1,0 +1,42 @@
+steps:
+
+- script: |
+    call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" $(python.architecture)
+    set MSSdk=1
+    set DISTUTILS_USE_SDK=1
+    python -m pip wheel . -w wheelhouse/
+  displayName: 'Build wheel (Windows Python 2.7)'
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'), eq(variables['python.version'], '2.7')) 
+
+- script: |
+    python -m pip wheel . -w wheelhouse/
+  displayName: 'Build wheel'
+  condition: and(succeeded(), not(and(eq(variables['Agent.OS'], 'Windows_NT'), eq(variables['python.version'], '2.7'))))
+
+- script: |
+    ls -lh wheelhouse
+    mkdir -p dist
+    cp wheelhouse/$(package_name)* dist/.
+  displayName: 'Show wheelhouse'
+
+- script: |
+    python -m pip install $(package_name) --no-index -f wheelhouse
+  displayName: 'Install wheel'
+
+- script: |
+    python -m pytest --pyargs iminuit --junitxml=junit/test-results.xml
+  # I'm selecting a random directory to not trigger relative imports.
+  workingDirectory: tutorial
+  displayName: 'Test with pytest'
+
+- task: PublishTestResults@2
+  inputs:
+    testResultsFiles: '**/test-*.xml'
+    testRunTitle: 'Publish test results for Python $(python.version)'
+  condition: succeededOrFailed()
+
+- script: |
+    python -m pip install delocate
+    /Library/Frameworks/Python.framework/Versions/$(python.version)/bin/delocate-wheel dist/$(package_name)*.whl
+  displayName: 'Delocate wheels'
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin')) 
--- a/.ci/azure-build-pipeline.yml
+++ b/.ci/azure-build-pipeline.yml
@@ -1,0 +1,104 @@
+trigger:
+  tags:
+    include:
+    - v*
+
+pr:
+- master
+
+
+# Required variables (set in UI):
+#   package_name:     This is the output name, - is replaced by _
+#   many_linux_base:  Should be quay.io/pypa/manylinux1_ or skhep/manylinuxgcc-
+variables:
+  package_name: iminuit
+  many_linux_base: quay.io/pypa/manylinux1_
+
+jobs:
+
+- job: LinuxSDist
+  pool:
+    vmImage: 'ubuntu-16.04'
+  variables:
+    python.architecture: 'none'
+  steps:
+    - script: |
+        python -m pip install setuptools cython
+        python setup.py sdist
+      displayName: Publish sdist
+    - template: azure-publish-dist.yml
+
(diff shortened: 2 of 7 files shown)


HDembinski:
Thank you very much for this patch, but I have two minor issues:

- This patch does some changes which seem unrelated to Azure (travis.yml, setup.py, setup.cfg). Could you please split off these changes into a separate PR? The changes look useful, but I can't make an exception now, since I was just a few days ago very strict with Matthias on that.

- Could you please use the existing folder `ci-scripts` to place the ci scripts instead of adding a new folder `.ci`?

HDembinski:
If you rebase to develop, the failing integration test will also work. Not necessary, but helpful to get rid of the irritating broken checks.

henryiii:
1. I'll pull out the changes into two other PRs. I originally made them to keep the CI scripts similar, but due to the fact that the tests are in a different folder *and* you can't run from the main directory with an installed iminuit (other binary package in Scikit-HEP have the same issue; but the workaround is different). I had to break compatibility with the previous scripts anyway because of this; so the changes, while still nice, are not tied to the current iteration of the pipeline.
2. Would you accept a PR to move `ci-scripts` to `.ci`? I would like to be consistent across Scikit-HEP, and I strongly prefer CI script directories to start with `.`, because they are not meant for users. However, the main reason is that the Azure pipeline script(s) have the ".ci" baked into it, and it would introduce a large inconsistency in these scripts. I am hoping to make many of these scripts shared across all Scikit-HEP at some point, to keep maintenance of the build system manageable. I can work around this if you really don't want to move the existing CI scripts.
3) (not in your list) Should I be compiling against an older Numpy?

.ci/azure-steps.yml
```diff
@@ -0,0 +1,42 @@
...
+    mkdir -p dist
+    cp wheelhouse/$(package_name)* dist/.
+  displayName: 'Show wheelhouse'
+
+- script: |
+    python -m pip install $(package_name) --no-index -f wheelhouse
+  displayName: 'Install wheel'
+
+- script: |
+    python -m pytest --pyargs iminuit --junitxml=junit/test-results.xml
```
henryiii: FYI, this is the block with the workaround that is not consistent with a general script for all packages. `workingDirectory` and `--pyargs iminuit`.

henryiii:
TODO:
- [ ] Add badge to readme
- [ ] Move folder if #344 is not accepted

I'm currently only planning for this to run on changes in master, rather than both master and develop, since it's meant for building final packages and not testing, but it's useful for testing (as we've found); it can be manually triggered on any *branch* easily (manual triggers on PRs are much harder or non-existent, though).

HDembinski:
> I strongly prefer CI script directories to start with `.`, because they are not meant for users.

Users pip install, everyone who checks out the source code is a potential developer. I like to look at the CI scripts in other projects.

HDembinski:
> (not in your list) Should I be compiling against an older Numpy?

I think it is fine like this.

henryiii:
CI scripts can change at any time, for example, every time Travis breaks something. The dot keeps people from assuming that this is safe to use. The 6K+ repositories in conda-forge, [for example](https://github.com/conda-forge/ablog-feedstock), use the `.` whenever they can. [Numpy](https://github.com/numpy/numpy) uses the dotted form for travis.yml and appveyor.yml, as does iMinuit itself. Why should the ci script directory be public when the ci main files themselves are not?

(Note: `azure-pipelines.yml` does not recognize a dotted version automatically; you have to set it in the GUI, so that one is currently often public)

henryiii:
And if a script is safe/meant for general consumption, it should not be in a CI folder.

HDembinski:
@henryiii I accepted the renaming of the ci folder, so this is the last PR to be include before the 1.3.6 release. What do I need to do to publish the wheels when I make the release?

HDembinski:
@henryiii Thank you very much for this patch and the Azure Tutorial, that is awesome and surely was a lot of work. I will quickly add the badge and then merge.

PR #341, opened 2019-05-18
fixing pypy3.5

HDembinski:


henryiii:
What about PyPy 7 support, could that be checked/added too? I know that PyBind11 doesn't support that one yet, don't know if it's a common issue or a big change, or just a small bug they happened to have; I don't really use PyPy.

HDembinski:
Python2.7 is dying, so I am not investing any energy in supporting the 2.7 branch of PyPy.

henryiii:
Ah, you did move to PyPy 7.1. I thought those were PyPy versions. +1

PR #342, opened 2019-05-18
Better pytest integration

henryiii:
This adds the [recommended method](https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner) of integrating PyTest with setup.py.

It also drops the redundant 'setuptools' from install_requires; setuptools must be available already for the file to run at all; PEP 518's `pyproject.toml` is the only way to change what happens before setup.py runs (and only for pip).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,5 @@
+[tool:pytest]
+testpaths=iminuit
+
+[aliases]
+test=pytest
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 # CFLAGS="-O0" python setup.py ... to override for debugging
 
 import os
+import sys
 import platform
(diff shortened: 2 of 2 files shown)


HDembinski:
Thank you! My setup.py-fu is not very good.

cdeil:
@henryiii @HDembinski - Do we need or want pytest integration in `setup.py` at all? Are there any advantages to have `python setup.py test` instead of just `pytest` for `iminuit`?

E.g. in Astropy and affiliated packages, most devs have been using just `pytest` for years, and probably `python setup.py test` will be removed soon.

HDembinski:
@cdeil I don't use it, but it is better if `python setup.py test` works now, even if it is removed soon.

PR #343, opened 2019-05-18
Travis cleanup

henryiii:
iMinuit is still using the pre-release Python 3.7 on Travis. This moves to the current architecture for Travis before Trusty is phased out. This is needed for 3.7 and the default [changes next week anyway](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment).

I'm also removing the double Travis build on PRs by checking for main branches only. This is already set up this way with Appveyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 
 cache:
   pip: true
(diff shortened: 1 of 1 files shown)


HDembinski:
Thank you for this! Just building master and develop is enough, no need to build version tags.

HDembinski:
It is weird that this patch is changing coverage. But I guess xenial comes with a different compiler, and I know that different versions of gcc produce different coverage data (newer versions are not always better),

PR #344, opened 2019-05-18
Moving the CI scripts

henryiii:
This would move `ci-scripts` to `.ci` if accepted.


HDembinski:
I am so-so on this. i don't see why the ci scripts should be hidden.

HDembinski:
Reading your argument in the other PR, I am with renaming to `.ci`.

henryiii:
I may have to make the azure support folder configurable somehow anyway, so only make the change if you do like it.

HDembinski:
@henryiii I am not strongly attached to the current folder name and I value consistency. I am ok with this PR if it helps to make scikit-hep projects more consistent.

Issue #348, opened 2019-05-23
Testing bugs


henryiii:
When making the wheels build, I ran into the following failing tests. They only fail for specific combinations of Python versions and hosts. You can see them [here](https://dev.azure.com/scikit-hep/iMinuit/_build/results?buildId=834), or pasted below.

The problems all seem to be that output is checked exactly using string comparison when it should be a bit looser. The problems are:

- [x] HTML attribute order is random
- [x] Extra 0's can be added in one of the matrix outputs
- [x] The color in LaTeX can be different by 1

### 64 bit: cp27-cp27mu

(seems to be somewhat random)

```
E       assert '<table>\n<tr...>\n</table>\n' == '<table>\n<tr>...>\n</table>\n'
E         Skipping 743 identical trailing characters in diff, use -v to show
E           <table>
E           <tr>
E         - <td colspan="2" title="Minimum value of function">
E         ?    ------------
E         + <td title="Minimum value of function" colspan="2">
E         ?                                      ++++++++++++...
E         
E         ...Full output truncated (56 lines hidden), use '-vv' to show

opt/_internal/cpython-3.5.7/lib/python3.5/site-packages/iminuit/tests/test_repr.py:53: AssertionError
```

```
E       assert '<table>\n<tr...>\n</table>\n' == '<table>\n<tr>...>\n</table>\n'
E         Skipping 43 identical leading characters in diff, use -v to show
E         Skipping 832 identical trailing characters in diff, use -v to show
E           </th>
E         - <td colspan="2" style="background-color:#92CCA6;" align="center">
E         ?     ^^ ^^^   ^                                     ---   ^^^^^^
E         + <td align="center" style="background-color:#92CCA6;" colspan="2">
E         ?     ^ ^^   ^^^^^^                                    +++++    ^...
E         
E         ...Full output truncated (71 lines hidden), use '-vv' to show

opt/_internal/cpython-3.5.7/lib/python3.5/site-packages/iminuit/tests/test_repr.py:413: AssertionError
```

### 32 bit linux

```
______________________________ test_latex_matrix _______________________________

    def test_latex_matrix():
        m = Minuit.from_array_func(Func9(), (0, 0), name=('x', 'y'),
                                   pedantic=False, print_level=0)
        m.migrad()
>       assert r"""%\usepackage[table]{xcolor} % include this for color
    %\usepackage{rotating} % include this for rotate header
    %\documentclass[xcolor=table]{beamer} % for beamer
[...]

PR #349, opened 2019-05-23
Wheel building updates

henryiii:
This changes the Azure backend, along with one small change to setup.py: it sets a minimum version of numpy to ensure that the user doesn't have a non-compatible numpy install.

Features of the new wheel building:
* Builds with an older version of Numpy (depending on Python version and what Numpy's were available when Python came out) - the minimum supported is now 1.11.3.
* Shareable bits are in a subdirectory, ready to become a new scikit-hep package `azure-wheel-helpers`.
* Any template can be overridden by a package by changing the `.ci/azure-pipeline-builds.yml`
* Tests are now run correctly and reported correctly for wheels (currently testing needs to be fixed before we can build manylinux wheels #348 )
* The `requirements-dev.txt` file was renamed to `.ci/requirement-build.txt`, since it's really not development requirements and it's really only for the azure wheel building.

The default test azure script now assumes a /tests folder, but can be (and is) overridden; the bash script used in the manylinux build, on the other hand, automatically detects and changes styles if the tests folder is not present.


HDembinski:
Please revert removing `requirements-dev.txt`, this is used by people who want to set up a development environment and it is documented.

HDembinski:
Also please trigger on changes to the develop branch.

HDembinski:
In fact, for my workflow it may be better to just trigger on develop and PRs, and leave the master. The master is always lagging behind develop, so it doesn't need extra checking.

HDembinski:
I don't understand what is wrong wit the travis test, it doesn't start

HDembinski:
Otherwise, the patch is now up to date. ManyLinux 32Bit is still failing, I know where the issue is coming from and I am working on it.

HDembinski:
It could be that the travis stuff only triggers when the PR is ready for review

henryiii:
The existing requirements-dev was added by me a few days ago (as dev-requirements) and was not really a developers requirements at all, but a list of things needed for the wheels (stripped down from requirements.txt). The current requirements.txt *is* what most people would call requirements-dev (it has Cython, for example), and could be renamed or copied and cleaned up.

henryiii:
I believe that Travis may not run on draft PRs, yes.

henryiii:
There is no need to have all wheels build every commit to develop (IMO). They really should only be built right before release, and maybe once in a while for testing major changes. If you want a stripped down test pipeline too, that’s easy to do (boost-histogram does that).

HDembinski:
> There is no need to have all wheels build every commit to develop (IMO).

We agree here. So once we have fixed the current issues with the builds, checks should only trigger on changes to the master.

> The existing requirements-dev was added by me a few days ago [...]

I am a bit fuzzy on how these files are supposed to be used, I am still learning about Python packaging. If `*requirements*.txt` is only used to set up the local developer environment anyway, then there should be only one such file called `requirements.txt`

henryiii:
@HDembinski 

> do we really need Linux 32bit

Probably not, but let's look at the download statistics; and I'd also really only recommend dropping 32-bit Linux when the Python Packaging Authority does (which may be during the switch to manylinux2010 - they are not botherinig with 32-bit, at least yet).

> It seems to me that manylinux2010 is sufficient.

I think you mean manylinux1? Manylinux2010 is brand new, and is only supported by relatively new pip versions ("not yet widely supported" is the tagline from the PPA). That is the bottleneck, not OS versions for hosts. I don't think anyone really has a good idea of when manylinux1 will no longer be needed; so far it requires Pip 10 or better. Pip 10 is *that* version of Pip, the one that completely removed the *entire* Python API, including `from pip import main`, which was "public" facing but not supposed to be public according to the Pip maintainers. This even broke the "pip" command itself on some platforms with no third-party modifications. Our download statistics will be *very* interesting. Once we see manylinux1 downloads drop to minimal amounts, we can drop them.

We can always drop supported platforms later, and add a note like "If you are using 32-bit Linux/pip9/Python 2, please install iminuit<2.0". I added the last one as inspirational. ;)

HDembinski:
Okay, keep manylinux1 but linux in 32bit should go. If a user has a weird system he/she can still use the sdist. We don't have to support weird platforms at the cost of Azure resources.

henryiii:
I believe we should follow the PPA's recommendations, as well as what other major projects are doing: Numpy builds 32-bit linux releases still: https://pypi.org/project/numpy/#files . I think this will be dropped soon (probably when the move to manylinux2010 finishes), but I'd rather drop it when PPA does. I believe Azure would rather a complete example at the cost of a *very* small amount of extra resources (it doesn't increase our total build time since it runs in parallel, and it only happens on changes to master, 1-2 times per release). If it was noticeably more maintenance burdon for us, I would drop it, but it shouldn't be.

I did design the support folder (soon to be repo) in such a way that every project selects exactly what platforms it wants to support.

PS: I don't have a way to select Python versions in manylinux, that's just the PPA provided Python versions minus 3.4 since Numpy doesn't build wheels for 3.4 either.

HDembinski:
Okay, all you say makes sense. I will not push this point.

PR #352, opened 2019-05-29
[WIP] Scikit-HEP wheel helpers - as git submodule may not be supported by Azure

henryiii:
This moves the azure-wheel-helpers into a shared repository for use with the rest of Scikit-HEP.

Azure may not allow you to load yaml files from submodules. WIP until I find a way around. If necessary, we can just vender this in manually (maybe with git tree?), but it would be nicer as a submodule.

Note that there are no code changes, just a few copyright statements and a README - should not block a release.


HDembinski:
Should I merge this?

henryiii:
No, unless Azure starts supporting submodules for CI files. I've been looking for a new way to do it.

PR #355, opened 2019-09-10
Sort input file list

bmwiedemann:
Sort input file list
so that `_libiminuit.so` builds in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ except ImportError:
     numpy_header = []
 
 libiminuit = Extension('iminuit._libiminuit',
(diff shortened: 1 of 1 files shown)


HDembinski:
If I understand correctly, this change may help to make all binaries from the same build environment identical, which is certainly a good thing. Thank you for the PR!

bmwiedemann:
Yes, that is the goal.
Thanks for merging.

Issue #368, opened 2019-11-17
user friendly setting of errordef


HDembinski:
As suggested several times now, we should have some simple constants in iminuit to set the errordef for ML and LSQ. Also requested was a simple way to set errordef so that one gets 1 sigma, 2 sigma, etc. errors.

Idea:
```
class ErrorDef:
    __slots__ = ["_scale", "_sigma"]
    def __init__(self, scale):
        self._scale = scale
        self._sigma = 1
    def __mul__(self, other):
        self._sigma *= other
        return self
    def __rmul__(self, other):
        return self.__mul__(other)
    def __float__(self):
        return self._scale * self._sigma ** 2
    def __eq__(self, other):
        return float(self) == other

ml_sigma = ErrorDef(0.5)
lsq_sigma = ErrorDef(1.0)

assert 2 * lsq_sigma == 4
```

PR #369, opened 2019-11-17
adding convenient setting of errordef

HDembinski:
Adds an easy way to set errordef in units of sigma or as confidence level (cl).
```py
    from iminuit import Minuit
    
    # this is for least-squares; for negative log-likelihood import from iminuit.errordef.nll 
    from iminuit.errordef.lsq import sigma, cl

    def lsq(mu):
        return (mu - 1) ** 2

    # for standard errors
    m = Minuit(lsq, mu=0, errordef=1 * sigma)
    m.migrad()
    # now m.errors["mu"] is approximately 1

    # for 2-sigma errors
    m.set_errordef(2 * sigma)  # or sigma * 2
    m.hesse()
    # now m.errors["mu"] is approximately 2

    # for 99 % confidence level
    m.set_errordef(0.99 * cl)  # or cl * 0.99
    m.hesse()
    # now m.errors["mu"] is approximately 2.6
```


HDembinski:
fixes #369

HDembinski:
@henryiii @jpivarski @eduardo-rodrigues @cdeil Any thoughts on this new feature? Easier setting of the error definition in iminuit was requested at the PyHEP workshop.

cdeil:
@HDembinski - I think you meant to x-ref #368

A previous discussion with my preference is here: https://github.com/scikit-hep/iminuit/issues/173#issuecomment-415688896

I think it's bad practice to play around with errordef, and to ever set it to anything but the canonical value for least square or likelihood. If `minuit.error["my_parameter"]` doesn't give me the 1 sigma error, my head explodes, I think this is too error prone to use and teach.

If you add this, then I strongly suggest to never suggest this:
```
# this is for least-squares; for negative log-likelihood import from iminuit.errordef.nll 
from iminuit.errordef.lsq import sigma, cl
```
Because then 200 lines down where `sigma` and `cl` is used, it's not visible what they are, whether for the nll or lsq case. That's a minor point though, you can fix that by changing to
```
from iminuit.errordef import lsq
```
and then using `lsq.sigma` and `lsq.cl` below.

If you could be persuaded to not introduce this dark magic of producing floats and fiddling with `errordef`, then there are still opportunities to make things simpler.

We could accept the string "lsq" or "nll" for "errordef" and it would choose the right value of 1 or 0.5.

We could teach how to convert back and forth between `sigma` and `cl` values via a `scipy.stats` call in the docs (it's always just a one liner, it doesn't need 100+ of lines of code as here). Or if we want to offer an API for that, we could have a utility function, or we could add a `cl` argument to functions that take a sigma like [minos](https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit.minos) and then the caller can choose what they like (and if they pass both sigma and cl we raise an error.

Overall I think this is a difficult API design choice, because it'll be used by most users, and from experience we know that this is hard to understand and teach by students and colleagues, and it's error-prone, one can easily get incorrect science results. It would be good to have as many people as possible chime in, although probably the 2 (or more) options to handle this would have to be clearly explained to get good feedback.

jpivarski:
I remember it was requested at the workshop, and I had uneasy feelings about it at the time. As someone else brought up later, sometimes likelihood is defined with an extra factor of 2.

When making a math library "simple," it needs to expose no more complexity than the underlying mathematics—in this case, users really should understand the details of their goal functions, which includes normalization. When someone asked, "How are we supposed to know that 0.5 means this and 1.0 means that?" I was thinking, "It's in Fred James's manual, which is more like a scientific article than software documentation."

But if there are a couple of standard values as strings, like `"lsq"` or `"nll"`, then I guess that's okay. It's a good way to confirm which one you've got without having to consult the manual. But I don't think this is a detail that should be "abstracted away," it's a detail to be reminded of at the appropriate time.

Speaking of the string values, I'd probably find `"lsq"` or `"nll"` as hard to remember as 0.5 and 1.0 (which, to be honest, I don't remember which is which). Isn't `"chi2"` an easier name? (I can't think of a better one for likelihood, other than the full word.)

What if the value of the `errordef` parameter (I'm elaborating on @cdeil's non-global state idea, not the global `import`) is one of three objects: `LeastSquares`, `Likelihood`, and `Other`, and the `__repr__` of each of these says its name and value—0.5, 1.0, or other—and setting the `errordef` parameter accepts numbers and strings and maps them to the appropriate object? That way, users can provide it in whatever way they remember and printing it back gives them confirmation (or refutation) of their choice.

This avoids treating the choice as an implementation detail, which it's not, but also helps to avoid mistakes.

cdeil:
> As someone else brought up later, sometimes likelihood is defined with an extra factor of 2.

I think it's quite common. E.g. in our software, we had this confusion which fit statistic functions did or didn't contain the factor 2, and chose to include the factor 2 for all of them (https://github.com/gammapy/gammapy/issues/2149), following the lead of earlier likelihood fitting codes in astronomy (https://cxc.harvard.edu/sherpa/statistics/).

> How are we supposed to know that 0.5 means this and 1.0 means that?

The main way to do it is via documentation.

Given that the MINUIT authors and http://lmu.web.psi.ch/docu/manuals/software_manuals/minuit2/mnerror.pdf specifically already called the errordef=1 and errordef=0.5 cases "chi2" and "likelihood", we could offer those strings as conveniences in addition to the current values that are accepted. Just put:
```
if errordef == "chi2":
    self.errordef = 1.0
elif errordef == "likelihood":
    self.errordef = 0.5
else:
    self.errordef = float(errordef)
```
But the documentation should still be improved, and now will need to explain that you should only choose "likelihood" if your likelihood doesn't contain the factor 2 already (i.e. it's `-log(L)`, not `-2*log(L)`, in any case it's not a likelihood. Actually, when writing this now, I think adding "likelihood" or "nll" isn't really adding clarify, overall it might be adding confusion.

My main point: I'm -100 on any scheme where `minuit.errors` doesn't yield one sigma errors. That will cause a lot of pain and confusion with users, if you give them more ways to do the same thing than we already have. I think the proposal here is such a scheme.

cdeil:
Following up on my suggestion above to only improve the documentation:

There is [this FAQ entry](https://iminuit.readthedocs.io/en/latest/faq.html#i-don-t-understand-minuit-hesse-minuit-minos-errordef-what-do-these-do) that basically covers this question of 

There should be one section or FAQ entry that explain the relation of `errordef` to the function normalisation, and also then the meaning of errors and `sigma`.

There is an FAQ entry on errordef at https://iminuit.readthedocs.io/en/latest/faq.html#i-don-t-understand-minuit-hesse-minuit-minos-errordef-what-do-these-do, that could be extended to explain how it works directly, keeping the link to the MINUIT manual for a longer explanation in addition.

This is pretty good: https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit.errordef
The formatting should be fixed and a link to the FAQ entry should be added.

This is also not bad: https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit.set_errordef
But it links to again some other PDF, instead it should link to the one high-quality FAQ entry in our docs.

This is pretty bad: https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit.minos
It doesn't mention errordef or doesn't mention how to compute `sigma` if the user wants some confidence level like 0.95. It could be improved by linking to the FAQ entry, to avoid explaining the same thing many times.

The most important place where people should set errordef is here: https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit
The explanation there isn't terrible, but it could be better (and should be identical to the other places of the errordef property and `set_errordef`):
> errordef: Optional. Amount of increase in fcn to be defined as 1 𝜎. If None is given, it will look at fcn.default_errordef(). If fcn.default_errordef() is not defined or not callable iminuit will give a warning and set errordef to 1. Default None(which means errordef=1 with a warning).

I didn't even know about this `fcn.default_errordef` option. Probably very few people are using this? Maybe we should discourage it, mentioning that we only keep it for backward compat? It's not clear how iminuit behaves if someone does this:
```
fcn.default_errordef = lambda _: 0.5
m = Minuit(fcn)
m.set_errordef(1.0)
```

Another place where a `sigma` option is offered:
https://iminuit.readthedocs.io/en/latest/reference.html#iminuit.Minuit.mncontour
Again with no good explanation, and actually a completely broken link: http://wwwasdoc.web.cern.ch/wwwasdoc/minuit/node7.html

Likelihood normalisation is intrinsically difficult to get right, given that there's different conventions in use. But I think apart from this intrinsic difficulty, my conclusion is that the problem is mainly the bad  `iminuit` documentation not explaining how it works in the right places where users look at docstrings and wonder what `errordef` or `sigma` they have to choose.

HDembinski:
Hi all, thank you for your inputs, but I must respectively disagree.

1) We don't want to use strings to set things. Because strings cannot be checked by the Python interpreter. We want to use enums or Python constants. If you mistype these, you get a syntax error. Something like that was my first idea, but people at the workshop also asked how to make 2 sigma errors etc. We want a solution that offers this. That only a few people will use this functionality is not a good argument, since the implementation cost is minor.
2) There legitimate reasons to want different errors than the default. You simply simply sometimes what 3 sigma errors or an interval with 99 % coverage.

HDembinski:
Some other points: There is no ambiguity about the likelihood normalisation. The factor 0.5 is correct when you write the likelihood as it is specified in every textbook.

HDembinski:
@cdeil I also don't dig your argument to let the user read the docs. The point of having a nice interface is make the code readable and to reduce the number of times necessary to read something up. Why do we make nice interfaces at all? We could all code stuff up in assembler, right? Just read up how it works.

HDembinski:
> I think it's bad practice to play around with errordef, and to ever set it to anything but the canonical value for least square or likelihood. If minuit.error["my_parameter"] doesn't give me the 1 sigma error, my head explodes, I think this is too error prone to use and teach.

Your head will not explode since you specifically set the errordef yourself when you construct the Minuit object. You always think in terms of a library that's using Minuit, and for those people nothing changes! But there are many people who use iminuit interactively in notebooks. iminuit has interactive in its name. And those people want to change the error definition now and then.

HDembinski:
@cdeil Thank you for pointing out where the docs are unclear, could you please add separate issues for these to issue tracker?

HDembinski:
> My main point: I'm -100 on any scheme where minuit.errors doesn't yield one sigma errors. That will cause a lot of pain and confusion with users, if you give them more ways to do the same thing than we already have. I think the proposal here is such a scheme.

You cannot simply deny that other users have such needs only because you don't.

jpivarski:
I think the danger of heads exploding is because there are two ways to ask for "N sigmas," through `errorder` and through some parameter in `minos` and `contours`. The former sets the scale for all deviations from the minimal point and the later sets the threshold (within that scale) for specific roots and level curves. The latter is what users usually want, since they might be making several curves on the same plot: e.g. 68%, 95%, 99%.

These two factors essentially multiply—there is a degeneracy between them. Personally, I tried to use the `minos`/`contours` setting as much as possible and `errordef` as little as possible. Since `errordef` is set at the time when the `Minuit` object is created, it's easy to forget what it was set to, and certainly bad practice to change that state between drawing each level curve. (For the normal reasons that stateful interfaces are bad.)

Therefore, I would only use it if I had to match an error definition built into the goal function. If, for instance, some people gave me correct likelihood functions and others gave me them with an extra factor of 2 (ostensibly to save me trouble, but actually compounding it), then I would use `errordef` to level them out. This never happened (I wrote my own goal functions, always on least squares problems) and so I never used it. That's why I have such difficulty remembering which is which.

The suggestion with strings was just to accept strings as a possible input, in addition to accepting numbers, and internally "normalizing" or "regularizing" them to a common format, probably enum-like objects. It's common in Python interfaces to accept a wide range of input types like a funnel, but it would be madness to pass around all those different types internally.

HDembinski:
> I think the danger of heads exploding is because there are two ways to ask for "N sigmas," through errorder and through some parameter in minos and contours

Ok, now I see the point. The problem is that iminuit was not designed with a strong focus on consistency and "make interfaces easy to use correctly and hard to use incorrectly". I am fighting these issues ever since I became maintainer.

Alright then, then I will paddle back on this. I will only offer something like an enum for easier setting of `errordef` and add a new keyword to `minos` and `mnprofile` to set the size of the interval as a confidence level. This actually has some advantages, too. 

Thank you for the discussion and finding this issue.

cdeil:
The original design of Minuit (see the documentation) is to use `errordef` only on `Minuit` construction, to declare how your likelihood function is normalised. Like I said, I suggest we stick with that and only improve the documentation, everything else will just create more confusion, not less. But I don't want to argue further, unsubscribing now. Good luck!

HDembinski:
> The original design of Minuit (see the documentation) is to use errordef only on Minuit construction, to declare how your likelihood function is normalised. Like I said, I suggest we stick with that and only improve the documentation, everything else will just create more confusion, not less. But I don't want to argue further, unsubscribing now. Good luck!

Unfortunately, someone added a public method to set errordef.

PR #371, opened 2019-12-21
Use latest azure-wheel-helper, add Python 3.8

henryiii:
This now uses git subtree to make this easier to maintain. The build system is being synced with the other Scikit-HEP projects.

This also adds Python 3.8.

We might consider adding optional PEP 517 builds and using Setuptools's built-in Cython support in the future. But not right now (PyJet/NumPythia are testing that first).


henryiii:
Note: I think this needs to be "normally" merged, as git rebase tends to break git subtree.

HDembinski:
Thank you for the update!

PR #398, opened 2020-05-04
fix: Use check-manifest v0.42 directory ignore pattern

matthewfeickert:
Use the new directory ignore pattern of `check-manifest` introduced in [`v0.42`](https://github.com/mgedmin/check-manifest/blob/master/CHANGES.rst#042-2020-05-03).

> You can ignore directories only by ignoring every file inside it. You can use `--ignore=dir/**` to do that

This fixes the issue in CI that has been seen in the [Azure Pipelines](https://dev.azure.com/scikit-hep/iMinuit/_build?definitionId=10&_a=summary) since the `check-manifest` `v0.42` release.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,14 +6,12 @@ test=pytest
 
 [check-manifest]
 ignore =
-    .ci
-    .ci/*
+    .ci/**
     .appveyor.yml
     .coveragerc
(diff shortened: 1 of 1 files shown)


HDembinski:
Thank you for this!

PR #401, opened 2020-05-05
Smarter setup

HDembinski:


matthewfeickert:
I'm not sure if the extra dependency is of interest to you, but `pyhf` uses [papermill](https://github.com/nteract/papermill) for [our notebooks tests](https://github.com/scikit-hep/pyhf/blob/master/tests/test_notebooks.py), which works quite well!

HDembinski:
> I'm not sure if the extra dependency is of interest to you, but `pyhf` uses [papermill](https://github.com/nteract/papermill) for [our notebooks tests](https://github.com/scikit-hep/pyhf/blob/master/tests/test_notebooks.py), which works quite well!

Thank you, this looks good, I will try papermill and look into pyhf for guidance.

PR #405, opened 2020-05-07
add print_level 3

HDembinski:


HDembinski:
@marinang For debugging you can now set the `Minuit.print_level` to 3, which produces more internal output. We didn't support this before, since it has a nasty side-effect, but if you need to do debugging, you probably don't care about it.

PR #467, opened 2020-07-25
fix: Allow subclasses to use from_array_func

henryiii:
Reported by @kratsg


henryiii:
I don't seem to be able to get rid of the stale "release check" check, but since this is going to develop now, it shouldn't matter.

HDembinski:
The failures are unrelated to the patch, so I am merging, thanks!

HDembinski:
Very strange that this triggered the CI jobs for PRs on master.

henryiii:
It pulled in history from the last PR, since it was the same commit. Why that one targeted master instead of develop, I don't know...

PR #471, opened 2020-07-30
Add numba to yml file

gonzaponte:
This package is a requisite. It is present in requirements-dev.txt, but not in environment-dev.yml.
It is needed by two tutorial tests to run.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -10,6 +10,7 @@ name: iminuit-dev
 dependencies:
   - python==3.6
   - numpy
+  - numba
   - scipy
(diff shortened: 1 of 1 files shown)


HDembinski:
No, this is correct. You don't need numba to use iminuit. The tutorials are not part of the iminuit software.

gonzaponte:
Ok. Should it be removed from requirements-dev.txt?

HDembinski:
No, because requirements-dev is for developers who want to run all unit tests, which includes tests of the notebooks.

Issue #477, opened 2020-09-11
Compatibility of parameter tables and similar with dark themes


l-althueser:
Hi iminuit team,

I noticed that some (or all) of the table and matrix representations using latex or html have fixed background colors for a clean and convenient coloring. However, when using dark themes in jupyter lab or vscode the fixed colors are a disadvantage. As a result, white text appears on a white colored table row.

I can image two possible fixes:

1. Removing the background colors or adding a kind of switch for dark or light themes.
2. Setting the font color explicitly in the html and latex outputs and preventing unreadable combinations (One probably also has to set a background color for currently transparent table cells).

As reference, I am talking about the colors set here: https://github.com/scikit-hep/iminuit/blob/develop/src/iminuit/repr_html.py#L7

Example for the current version:
![image](https://user-images.githubusercontent.com/13698969/92901011-72c70d80-f420-11ea-836d-7d6ad7bd8c28.png)

.. if one monkey patches the colors:
![image](https://user-images.githubusercontent.com/13698969/92901101-84101a00-f420-11ea-9b22-beadb1b6e8e2.png)

```
import iminuit
iminuit.repr_html.backgrounds = ("background-color:#252526;", "background-color:#1E1E1E;")
iminuit.repr_html.good_style = "background-color:#415944;"
iminuit.repr_html.bad_style = "background-color:#B17E6A;"
iminuit.repr_html.warn_style = "background-color:#D2D2A4;"
```

Thank you!

HDembinski:
If there is an automatic way to pick up the surrounding style, I would like to implement it, but adding color themes is beyond the scope of iminuit. However, I have no problem adding a section in the docs that demonstrates this monkey-patching. Are interested in adding that to the docs?

PR #478, opened 2020-09-11
Add font style to colored html table cells

l-althueser:
I would propose the following minimal "fix" for iminuit output in dark themed environments. I won't expected interference with any current jupyter usage. A black font color was added explicitly for cells with an explicit background color to force readability independent on the exact environment.
Transparent cells will still be colored (both, text and background) with respect to the user environment.

Here is an example for the proposed changes in a dark themed environment (light themed looks identical to what most user are used to):
image

Feel free to reject this pull request if this is not the way you want to address the "issue" (see #477 for more info).
Thanks.

### DIFF
--- a/src/iminuit/repr_html.py
+++ b/src/iminuit/repr_html.py
@@ -1,10 +1,10 @@
 from iminuit.color import Gradient
 from iminuit.repr_text import pdg_format, matrix_format, fmin_fields
 
-good_style = "background-color:#92CCA6;"
-bad_style = "background-color:#c15ef7;"
-warn_style = "background-color:#FFF79A;"
-backgrounds = ("background-color:#F4F4F4;", "background-color:#FFFFFF;")
+good_style = "background-color:#92CCA6;color:#000;"
+bad_style = "background-color:#c15ef7;color:#000;"
+warn_style = "background-color:#FFF79A;color:#000;"
+backgrounds = ("background-color:#F4F4F4;color:#000;", "background-color:#FFFFFF;color:#000;")
 
 
 def to_str(tag):
--- a/src/iminuit/tests/test_repr.py
+++ b/src/iminuit/tests/test_repr.py
@@ -141,20 +141,20 @@ def test_html_fmin_good(fmin_good):
         <td colspan="3" style="text-align:center" title="Increase in FCN which corresponds to 1 standard deviation"> up = 0.5 </td>
     </tr>
     <tr>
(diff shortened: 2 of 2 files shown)


HDembinski:
Hi, sorry for the long delay, this looks like a nice improvement. Also very nice to show a screenshot!

I am not a CSS expert, so I don't really understand what adding `;color:#000` does.

The patch seems to fix the issue only half-way. The value table is still bright. Can this be fixed as well?

l-althueser:
Hi! 
Adding `;color:#000` will force the text color of all colored table cells to be black. In the example screenshot above and in #477 is the table shown with a dark color theme. Usually programs will "just" flip colors and change the default text color to white if not explicitly defined.
Previous to this PR, iminuit was only setting the table cell color explicitly, assuming the text color to be always black. But in dark themed environments, the text color is white by default - so the user would see white text over white value cells. 

The dark colored part of the tables is just a result of the usual dark theme and not influenced by the PR. So I went for a minimal change in iminuit to keep the usual colors that all users know while fixing the issues with dark themes that not many users will experience.

HDembinski:
Thank you for the great explanation. After also reading your issue, I fully understand this now. Forcing the text color to black in the function minimum block makes sense. For the table of variables, ideally we would like to have something that alternates between "make current background a bit lighter" and "make current background a bit darker". Perhaps there is way to express this?

l-althueser:
I don't think that there is an easy way to detect the color theme that the user is using (I might be wrong) given that there are many ways to run iminuit. In principle one could start and adding more complex CSS styles that automatically switches colors and so on. But I don't think that the community of iminuit will mind having a white table if they are able to see the fit result ;)

So actually I am also not an expert on CSS styling python output for jupyter .. An expert might see the table in the future and having a nice idea to solve it automatically. Until then, I would be in favor of adding this PR to fix the issue even if the styling is not optimal for dark themes.
Thanks!

HDembinski:
It looks like tables are automatically styled with alternating light/dark shades, so removing the manual coloring fixes the variables table. I am going to close this PR in favor of another that is in preparation, which fixes this and other coloring issues. Thank you!

l-althueser:
That might help for the variable table! I think it would still be good for the green/red colored cells to set the text color explicitly. Thanks

HDembinski:
> I think it would still be good for the green/red colored cells to set the text color explicitly.

Sure, I got that. I just release 1.5.0 with your patch included.

PR #481, opened 2020-09-17
dark theme friendlier html style

### LINKED ISSUE #477: Compatibility of parameter tables and similar with dark themes

HDembinski:
Fixes #477

### DIFF
--- a/src/iminuit/repr_html.py
+++ b/src/iminuit/repr_html.py
@@ -1,10 +1,9 @@
 from iminuit.color import Gradient
 from iminuit.repr_text import pdg_format, matrix_format, fmin_fields
 
-good_style = "background-color:#92CCA6;"
-bad_style = "background-color:#c15ef7;"
-warn_style = "background-color:#FFF79A;"
-backgrounds = ("background-color:#F4F4F4;", "background-color:#FFFFFF;")
+good_style = "background-color:#92CCA6;color:black"
+bad_style = "background-color:#c15ef7;color:black"
+warn_style = "background-color:#FFF79A;color:black"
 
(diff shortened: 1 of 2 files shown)


l-althueser:
Thank you!

Just as reference, this is the result:
image

HDembinski:
I tried this on my computer with a dark jupytertheme and every second row was highlighted. Is this not the case for you?

l-althueser:
This is probably a _feature_ of visual studio code. I do get a highlighted row on mouseover so that I would say the implementation is fine. Thanks

PR #488, opened 2020-09-24
make iterate configurable

HDembinski:
@matthewfeickert @kratsg This patch makes the iteration ("try hard") feature of iminuit configurable. You should definitely try to play with the precision keyword, too, because I did and it improve convergence when the cost function has low precision (the doc string was extended to better explain what this setting does).

### DIFF
--- a/src/iminuit/_libiminuit.pyx
+++ b/src/iminuit/_libiminuit.pyx
@@ -795,7 +795,7 @@ cdef class Minuit:
         return cls(fcn, **kwds)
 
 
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> This patch makes the iteration ("try hard") feature of iminuit configurable. You should definitely try to play with the precision keyword, too, because I did and it improve convergence when the cost function has low precision (the doc string was extended to better explain what this setting does).

Fantastic! Many thanks @HDembinski and we'll get on that (I'll make a follow up Issue on `pyhf`).

PR #525, opened 2020-11-16
style: pyupgrade

henryiii:
Trying out PyUpgrade to clean up older syntax to cleaner 3.6+. Idea from @kratsg & @matthewfeikert in https://github.com/scikit-hep/pyhf/pull/1164 .

Only the first commit is hand written - the second commit is entirely generated.

- style: add pyupgrade
- style: pre-commit run -a

### DIFF
--- a/.ci/install_pypy36.py
+++ b/.ci/install_pypy36.py
@@ -10,7 +10,7 @@ if Path("pypy36/pypy3/bin/activate").exists():
     raise SystemExit
 
 plat = {"darwin": "osx64", "linux": "linux64"}[sys.platform]
-filename = "pypy3.6-v7.3.1-{}.tar.bz2".format(plat)
+filename = f"pypy3.6-v7.3.1-{plat}.tar.bz2"
 
 url = "https://bitbucket.org/pypy/pypy/downloads/"
 plat = {"darwin": "osx64", "linux": "linux64"}[sys.platform]
--- a/src/iminuit/_minuit.py
+++ b/src/iminuit/_minuit.py
@@ -793,7 +793,7 @@ class Minuit:
 
         if not self._fmin.is_valid:
             raise RuntimeError(
-                ("Function minimum is not valid. Make sure " "MIGRAD converged first")
+                "Function minimum is not valid. Make sure " "MIGRAD converged first"
             )
         if var is not None and var not in self._pos2var:
             raise RuntimeError(f"Unknown parameter {var}")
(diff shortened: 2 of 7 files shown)


src/iminuit/repr_html.py
```diff
@@ -30,13 +30,13 @@ def tag(name, *args, **kwargs):
...
         v = kwargs[k]
-        head += ' %s="%s"' % (k, v)
+        head += f' {k}="{v}"'
     head += ">"
     tail = "</%s>" % name
     if len(args) == 0:
         return [head + tail]
     if len(args) == 1 and isinstance(args[0], str):
-        return ["{0} {1} {2}".format(head, args[0], tail)]
+        return ["{} {} {}".format(head, args[0], tail)]
```
henryiii, line 39: pyupgrade tried to be very fugal with f-strings, only using them when it's clearly better - if there are non-simple expressions, like `args[0]`, it's won't automatically suggest they be used (though some of those cases, I still think the f-string would be better).

.ci/install_pypy36.py
```diff
@@ -10,7 +10,7 @@
     raise SystemExit
 
 plat = {"darwin": "osx64", "linux": "linux64"}[sys.platform]
-filename = "pypy3.6-v7.3.1-{}.tar.bz2".format(plat)
+filename = f"pypy3.6-v7.3.1-{plat}.tar.bz2"
```
henryiii, line 13: Is this file still needed? GHA has PyPy support built in, as does cibuildwheel.

src/iminuit/_minuit.py
```diff
@@ -1452,9 +1454,9 @@ def __eq__(self, other):
         return len(self) == len(other) and all(x == y for x, y in zip(self, other))
 
     def __repr__(self):
-        s = "<%s of Minuit at %x>" % (self.__class__.__name__, id(self._minuit))
+        s = "<{} of Minuit at {:x}>".format(self.__class__.__name__, id(self._minuit))
```
henryiii, line 1457: Very nice that it can replace old-style formatting - I wonder if it has a 2.7 mode I can apply to very old projects? :)

.ci/install_pypy36.py
HDembinski: Probably not.

HDembinski:
Looks all good so will merge, but I think stylistics upgrades are up the maintainer and shouldn't be commit to "try out new tools".

henryiii:
You can always ignore or change it - and that's why I even made the branch in a fork, so there'd be no history at all if you didn't like it. Anyone can suggest a change on GitHub, that's kind of the point - the active maintainer should make the final decision and merge or close, but anyone can suggest. Even if you didn't want to merge it, it's still interesting to see what it does.

HDembinski:
You discuss a technical issue, I am discussing a social issue.

PR #527, opened 2020-11-16
fix: avoid pyproject's default __legacy__ behavior

henryiii:
While stealing the config for https://github.com/scikit-hep/hist/pull/119, I noticed that the `build-backend` key is missing from `pyproject.toml`.
This causes the `"setuptools.__legacy__"` backend to be selected, which is
generally incorrect for a normal build. For example, if you use PyPA-build,
this may fail with `__legacy__` selected. This add the recommended selection of
`"setuptools.build_meta"`.


henryiii:
I've been noticing that there is an "import cmake" from setup.py, and I had no idea why that was working, importing from a local directory is not supposed to work in setup.py. It turns out it was depending on the `__legacy__` behavior which put `.` at the beginning of sys.path, even overriding the `cmake` module that is installed with CMake.

I made the import explicit, and also changed the name of `cmake.py` to not collide with `cmake` the module that we install in pyproject.toml (@HDembinski I just randomly chose a name, feel free to change or ask me to change).

HDembinski:
Looks good, thank you and also for the helpful explanation. Now that you pointed it out, I am also surprised that `import cmake` worked...

PR #536, opened 2020-11-20
Pretty printing of minuit

HDembinski:


HDembinski:
To-do: better repr for FMin, MError etc.

PR #540, opened 2020-11-22
added sphere_np to test functions; test repr_pretty of Minuit and others

HDembinski:


HDembinski:
Minuit.reset now also returns self

PR #569, opened 2020-12-08
tests: include tests in SDist

henryiii:
The SDist does not contain the tests. While the _wheel_ should not contain tests (and I didn't check to verify that by adding it, `/tests` is not picked up as a package or as package data, which you should do), it is helpful if the SDist contains tests, so you can build and then verify the build worked, say in conda-forge. I'll look at a different way to solve this on conda-forge, but this is a suggestion.

Edit: from a very quick check, it looks fine. It's rare to have `/src/<package>` based packages accidentally pick up `/tests` in the wheel.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,8 @@
 include src/*.cpp
 include src/*.hpp
 include src/iminuit/*.py
+include tests/*.py
+include tests/*.txt
 recursive-include extern/root/math/minuit2/inc *.h
 recursive-include extern/root/math/minuit2/src *.cxx
 recursive-include extern/pybind11/include *.h


HDembinski:
I think the tests should not be in the sdist if they are not in the wheels. In the conda build you should be able to assume that the sdist is build correctly if it builds, because I am testing the sdist before release. As a test that everything was built correctly, a simple import iminuit should then be enough.

henryiii:
Unless the tests take a long time to run(1), it could be helpful to include them so that anyone building from SDist can check to make sure their code runs. For example, conda-forge builds ARM, Apple Silicon, PyPy - unless you test all those things, it might expose a bug in the binaries you are creating in conda-forge. Also the compiler settings, etc. are all different. Also if someone builds a binary on Alpine or ClearLinux, it can be helpful to verify the binary works. For wheels, those are already built and vetted (tests run on them usually), so there's no reason to include the tests in the wheels / importable package, in my opinion. However, most really large projects (NumPy, Pandas, etc) do what iMinuit used to do, and make the tests available directly in the final package, because it's so critical to be able to verify successful builds by packagers in various systems. I'd rather leave the production systems that get wheels light and lean, but that's a choice.

SDists do not have to only contain exactly what goes into wheels. For example, if your tests are a bit large, there is only one SDist per release, but many wheels, so you aren't storing many copies of the test. Also, you don't have to put the whole test suite in, just enough to be able to say "yes, everything should work" if someone builds a binary.

If you don't think it's a good idea, we could stick to `import iminuit`, or use the current workaround in the conda-forge recipe. PyHF's feedstock, which probably has very slow tests, just does an import and a version check. NumPy's feedstock runs the tests.

1: If the tests take a long time to run, really all that's needed here is a small, minimal test to make sure things built correctly, so just a single file or two might be just fine, or maybe a simple recipe we could drop into the iminuit-feedstock repo?

HDembinski:
You make very good points, @henryiii. Sorry for taking so long, I merged it.

Issue #570, opened 2020-12-08
Some versions of MSVC not supported


henryiii:
The following:

https://github.com/scikit-hep/iminuit/blob/f12cd519aa08e244ad71f56fe7a97d42117196e4/cmake_ext.py#L37

is not supported by conda-forge, I get:

```
  CMake Error at CMakeLists.txt:3 (project):
    Generator
      Visual Studio 15 2017 Win64
    does not support platform specification, but platform
      x64
    was specified.
```

The problem is that Conda-forge sets the generator for you, so trying to assume and set things breaks it. If you check for and avoid setting this if `CMAKE_GENERATOR` is set, then I think this will work.

See the `contains_arch` code here: https://github.com/pybind/cmake_example/blob/master/setup.py

HDembinski:
Thanks, I will have a look into this.

henryiii:
Gentle ping - I can't progress on the conda-forge packaging until this supports the conda forge compilers. :)

henryiii:
Sorry, I forgot to close this. Thanks!

PR #576, opened 2020-12-14
added msvc fixes from pybind11/cmake_example by henryiii

HDembinski:


HDembinski:
Closes #570

henryiii:
You need to remove the `def run(self):...` method. It's not initializing the compiler (which is why self.compiler is empty). The default run method is correct - I fixed this bug in the old example when I rewrote it: https://github.com/pybind/cmake_example/blob/580c5fd29d4651db99d8874714b07c0c49a53f8a/setup.py#L39

henryiii:
I can apply this as a patch to see if it fixes conda without having to make a release, so we don’t get stuck in a loop if there is still a problem. Will apply soon.

henryiii:
Thanks for keeping the CI changes separate, I would have had to manually modify the patch (I found out because I accidentally applied the patch from my branch instead of this one).

henryiii:
I can verify that this patch fixed conda-forge! A patched 2.0.0 is released. Thanks for including `CMAKE_BUILD_PARALLEL_LEVEL` too, as that let me fix the out-of-memory error on Travis as well.

henryiii:
We (you, me, or @chrisburr, probably) need to remember to simplify the recipe when 2.0.1 or 2.1.0 comes out, dropping this patch and the manual inclusion of the tests via git.

HDembinski:
Thank you very much for helping with this and having the eye on conda-forge @henryiii

PR #577, opened 2020-12-15
Fix/conda2

henryiii:
Removes the overload to `run`, which was missing the setup steps, including initializing the compiler. Setuptools tries to copy the created final .so file from the build directory to the inplace directory when it's done - this was broken by the run not initializing correctly - so now the extension has to have the expected name - and if you use the full name, `iminuit._core`, setuptools handles the folder for you.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ name: CI
 on:
   pull_request:
     branches:
+      - develop
     paths-ignore:
       - 'doc/**'
       - '.ci/**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,6 @@ name: CI
 on:
   pull_request:
     branches:
-      - develop
     paths-ignore:
       - 'doc/**'
       - '.ci/**'
--- a/cmake_ext.py
+++ b/cmake_ext.py
@@ -7,18 +7,14 @@ from pathlib import Path
 
 
 class CMakeExtension(Extension):
-    def __init__(self, sourcedir=""):
-        Extension.__init__(self, "CMakeExtension", sources=[])
+    def __init__(self, name, sourcedir=""):
+        Extension.__init__(self, name, sources=[])
         self.sourcedir = os.path.abspath(sourcedir)
 
 
 class CMakeBuild(build_ext):
-    def run(self):
-        for ext in self.extensions:
-            self.build_extension(ext)
-
     def build_extension(self, ext):
-        extdir = Path(self.get_ext_fullpath(ext.name)).parent.absolute() / "iminuit"
+        extdir = Path(self.get_ext_fullpath(ext.name)).parent.absolute()
         # required for auto-detection of auxiliary "native" libs
         cmake_args = [
             f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}/",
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -3,7 +3,6 @@ name: CI
 on:
   pull_request:
     branches:
-      - develop
```
henryiii, line 6: Optional, but making a PR into a PR (like this one) is handy.

cmake_ext.py
```diff
@@ -7,18 +7,14 @@
 
 
 class CMakeExtension(Extension):
-    def __init__(self, sourcedir=""):
-        Extension.__init__(self, "CMakeExtension", sources=[])
```
henryiii, line 11: This was causing setuptools to look for `CMakeExtension.*.so`. This means this technique will not work if you produce multiple "Extensions" from a single "CMakeExtension" - but at some point we are really just recreating scikit-build from scratch.

cmake_ext.py
```diff
@@ -7,18 +7,14 @@
...
 
 
 class CMakeBuild(build_ext):
-    def run(self):
-        for ext in self.extensions:
-            self.build_extension(ext)
-
     def build_extension(self, ext):
-        extdir = Path(self.get_ext_fullpath(ext.name)).parent.absolute() / "iminuit"
+        extdir = Path(self.get_ext_fullpath(ext.name)).parent.absolute()
```
henryiii, line 17: If you include a dot in the name, you don't have to fiddle with the path here. And it works in the check functions too!

cmake_ext.py
```diff
@@ -35,7 +31,7 @@ def build_extension(self, ext):
         cmake_args += [f"-DCMAKE_BUILD_TYPE={cfg}"]
         build_args = ["--config", cfg]  # needed by some generators, e.g. on Windows
 
-        if self.compiler and self.compiler.compiler_type == "msvc":
+        if self.compiler.compiler_type == "msvc":
```
henryiii, line 34: If `self.compiler` is not set, you are already in big trouble...

henryiii:
Okay, @HDembinski, ready for review. This is a PR into your branch, not master/main/develop/whatever. Feel free to edit as you see fit. Or you could make a PR into my PR into your PR... ➿

cmake_ext.py
HDembinski: This was there because it is needed.

cmake_ext.py
HDembinski: Agreed, this used to be simple enough to have a custom implementation, but not anymore.

.github/workflows/ci.yml
HDembinski: Can't you just make suggested edits on the PR?

HDembinski:
@henryiii Thank you for the fixes, I hope this resolves the conda issues?

HDembinski:
Nice trick with the branch of the branch.

.github/workflows/ci.yml
henryiii: Yes, if you look at the history of your PR you will see a force push from me that seems to do nothing. It turned out to a bit less simple than I expected to fix (I didn’t catch the name issue at first), so I undid that and moved to a PR.

henryiii: Oh, suggested edits, no, you can’t suggest an edit for something more than a few lines away from a change. And you can’t see if the edit passed CI. :)

PR #601, opened 2021-01-23
chore: bump cibuildwheel to 1.8.0

henryiii:
Updating to the latest cibuildwheel, and using dependabot to keep this up to date in the future (it makes a PR if there's an update once a week). Also simplified a few things, fixed the only-binary setting, and added a workflow dispatch trigger so this can be manually tested. If it can't be, you might want to commit the workflow dispach line only to master, then manually trigger it on this PR. Please trigger a build before merging!

Other PRs in the migration here: scikit-hep/scikit-hep.github.io#112


.github/workflows/wheels.yml
```diff
@@ -19,23 +20,18 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-20.04, windows-latest, macos-latest]
```
henryiii, line 23: By the way, you probably should pin all or use latest on all, rather than a mix. This likely was from an example in cibw that was like this but is now consistent.

HDembinski:
Why do I need a dependabot for cibuildwheels? I don't see any merit in that.

HDembinski:
I don't want to depend build which is version 0.1.

HDembinski:
Note that these tests do not actually test cibuildwheels. I can test and merge these changes when I make a release.

henryiii:
> Why do I need a dependabot for cibuildwheels?

Why would you want to have an old version of cibuildwheel? It should be pinned, we are constantly fixing things like issues with CI system updates, PyPy problems, etc, and tracking the latest manylinux images. There is also an extremely  important update coming soon that will enable building Apple Silicon wheels. I'd prefer to have a bot make an update once a week if there's a cibuildwheel update available rather than having to go to every Scikit-HEP repository and bump it manually like I'm doing now. Automation is much better than manual, as long as there's a manual accept/deny step.

> I don't want to depend build which is version 0.1.

That's fine, but it's built on top of mature tooling and is an official PyPA project. It also correctly resolves PEP 518 dependencies while manually calling setup.py cannot.

HDembinski:
> > Why do I need a dependabot for cibuildwheels?
> 
> Why would you want to have an old version of cibuildwheel? It should be pinned, we are constantly fixing things like issues with CI system updates, PyPy problems, etc, and tracking the latest manylinux images. There is also an extremely important update coming soon that will enable building Apple Silicon wheels. I'd prefer to have a bot make an update once a week if there's a cibuildwheel update available rather than having to go to every Scikit-HEP repository and bump it manually like I'm doing now. Automation is much better than manual, as long as there's a manual accept/deny step.

Because my setup works for me.

> > I don't want to depend build which is version 0.1.
> 
> That's fine, but it's built on top of mature tooling and is an official PyPA project. It also correctly resolves PEP 518 dependencies while manually calling setup.py cannot.

The status quo works for me, so no reason to change.

henryiii:
> I can test and merge these changes when I make a release.

I know, no rush, and please make sure it's working before tagging anything. :) Added `workflow_dispatch` will give you a button to push to manually test in the future.

.github/workflows/wheels.yml
```diff
@@ -10,7 +10,7 @@ on:
     types: [published]
 
 env:
-  CIBW_ENVIRONMENT: "PIP_ONLY_BINARY=:all:"
+  CIBW_BEFORE_BUILD: "python -m pip install numpy --only-binary=:all:"
```
henryiii: Why did you revert this? You don't need NumPy to build since you have built on pybind11, so you either need `CIBW_BEFORE_TEST` or better yet, `CIBW_ENVIRONMENT`.

HDembinski: Perhaps you are right, perhaps you are wrong, but setting this up took a lot of time and I am not interested in touching this.

HDembinski: Unless something is broken, please don't "fix" it in my repositories.

henryiii: Thanks for the vote of confidence.

henryiii: I have not touched anything in `github.com:hdembinski`. I am responsible for maintaining the build systems for Scikit-HEP, and am trying to work on enabling our packages to be used in as many places as possible, such as on PyPy, ARM, and Apple Silicon. If you tie NumPy to the build step, then you will not be able to support builds where NumPy is not available as a wheel without building it yourself (which is really slow on ARM/PyPy, and hard on Apple).

As lead maintainer, it is your decision to accept anything I suggest, or not. But please assume I have some reason for suggesting changes. If you did not want Scikit-HEP maintainers trying to help maintain iMinuit, you shouldn't have moved it to Scikit-HEP.

HDembinski:
Merging into beta for testing

Issue #608, opened 2021-02-04
util.describe fails for certain functools.partial functions


jnsdrtlf:
I recently came across an issue were I tried to use [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial) to simplify my function. This error only happened when using `*args` or `**kwargs` in the partial function.

### Quick Example

This might not be a very practical example, as the minimum is rather trivial, but one could construct a reasonably difficult function that works the same way. If you are interested in a more real-world example, I'd be happy to share my code which originally lead to this issue.

```python
import numpy as np
from iminuit import Minuit
from functools import partial


def foo(x, *args):
    arguments = np.array(args)
    return np.sum(x - arguments, axis=0)


p_foo = partial(foo, 42)  # x = 42


# Minimum is trivial (should be at args[0] = 0.0 and args[1] = 0.0)
m = Minuit(p_foo, 0.04, 0.02)  # Fails!
```

More specifically, the error is:
```text
AttributeError: 'functools.partial' object has no attribute '__name__'
```
caused by this line:
https://github.com/scikit-hep/iminuit/blob/2b7b4a815629a8ebc97109a2e46887b8139c7715/src/iminuit/util.py#L877

### Cause of this Issue

When the keyword argument `name` is not passed to the `Minuit` constructor, the method [`iminuit.util.describe(callable)`](https://github.com/scikit-hep/iminuit/blob/2b7b4a815629a8ebc97109a2e46887b8139c7715/src/iminuit/util.py#L773) is used to determine the argument names from the functions signature.
In this function, we have the following three functions that help us to retrieve the argument names:

https://github.com/scikit-hep/iminuit/blob/2b7b4a815629a8ebc97109a2e46887b8139c7715/src/iminuit/util.py#L833-L835

The first will fail as a `partial` object has no attribute `func_code`. The second (`_arguments_from_inspect`) usually works fine with `partial` objects as `inspect.signature` returns the correct signature. In this case this would be `("*args",)`. However, `*args` and `**kwargs` are ignored as for these types, we construct our own argument names (`x0`, `x1`, ...). So we will eventually end up in the last function `_arguments_from_docstring` which will fail because there is no attribute `__name__` associated with a partial object. Thus, an exception is raised.

Note that this is not solved by using the `func` attribute that is provided by the partial object. This whole function is not going to work with partial functions as `inspect.getdoc` will always return the following:
```text
[...]

PR #609, opened 2021-02-06
Fix iminuit.util.describe(callable) for partial functions without signature

jnsdrtlf:
More information on this can be found in issue #608.

I'm happy for any feedback regarding the fix :)

Added some tests in both `tests/test_describe.py` as well as `tests/test_minuit.py`.

### DIFF
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -874,7 +874,12 @@ def _arguments_from_docstring(obj):
     #   min(arg1, arg2, *args, *[, key=func]) -> value
     #   Foo.bar(self, int ncall_me =10000, [resume=True, int nsplit=1])
 
-    name = obj.__name__
+    try:
+        name = obj.__name__
+    except AttributeError:
+        # See Issue #608
+        # https://github.com/scikit-hep/iminuit/issues/608
+        return ()
 
     token = name + "("
     start = doc.find(token)
--- a/tests/test_minuit.py
+++ b/tests/test_minuit.py
@@ -204,6 +204,7 @@ def test_Func2():
 
 def test_no_signature():
     from functools import partial
+
     def no_signature(*args):
         x, y = args
         return (x - 1) ** 2 + (y - 2) ** 2
--- a/tests/test_minuit.py
+++ b/tests/test_minuit.py
@@ -203,6 +203,7 @@ def test_Func2():
 
 
 def test_no_signature():
+    from functools import partial
     def no_signature(*args):
         x, y = args
         return (x - 1) ** 2 + (y - 2) ** 2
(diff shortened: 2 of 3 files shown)


tests/test_minuit.py
```diff
@@ -213,6 +215,17 @@ def no_signature(*args):
...
+    m = Minuit(partial_no_signature, 3)
+    assert m.values == (3,)
+    assert m.parameters == ("x0",)
+    m.migrad()
+
+    m = Minuit(lambda *args: no_signature(42, *args), 2)
+    assert m.values == (2,)
+    assert m.parameters == ("x0",)
+    m.migrad()
+
```
HDembinski, line 228: Is this second test really necessary? It seems to me that test_generic_partial() is sufficient. Am I missing something?

src/iminuit/util.py
```diff
@@ -874,7 +874,12 @@ def _arguments_from_docstring(obj):
     #   min(arg1, arg2, *args, *[, key=func]) -> value
     #   Foo.bar(self, int ncall_me =10000, [resume=True, int nsplit=1])
 
-    name = obj.__name__
+    try:
+        name = obj.__name__
+    except AttributeError:
+        # See Issue #608
+        # https://github.com/scikit-hep/iminuit/issues/608
+        return ()
```
HDembinski, line 882: I think that is fine for now, but I wonder whether it is possible to get the function name in a more generic way, perhaps with the inspect module so that it also works with a wrapped function.

HDembinski (commented):
I would remove the seemingly superfluous tests, otherwise this is a fine patch that I am happy to accept. I am going to wait a bit for your reply but will merge this by the end of the week if I don't hear from you.

Thanks!

tests/test_minuit.py
jnsdrtlf: It probably is. I first added this test, as the lines above and below already suggested a similar scenario and later realized that `tests/test_describe.py` is probably more fitting. I can remove it, sure. But I think it is also no problem if we keep it here. What do you think?

src/iminuit/util.py
jnsdrtlf: It is! Partial objects have an attribute `partial.func`. This attribute basically holds the function object itself.

However, once you go down this rabbit hole, you will eventually end up writing a completely new function just for partial objects. Keep in mind that `_arguments_from_inspect` **already works perfectly fine for partial objects**. The only problem is that `_arguments_from_inspect` returns `()` (and rightfully so) because it only finds `*args` which it ignores. Thus, we execute the next function in the chain which is `_arguments_from_docstring` which simply does not work for partial objects.

jnsdrtlf:
Thank you for the quick feedback. I can remove the unnecessary test and commit again :) Yay, this will mark my first contribution to a scikit-hep project! How exciting

src/iminuit/util.py
HDembinski: Ok, let's keep it like this for now.

tests/test_minuit.py
HDembinski: I prefer to remove these extra tests and only keep the test_generic_partial, then. Could you make the change?

jnsdrtlf: Done!

HDembinski:
Thanks for the patch!

PR #617, opened 2021-03-06
Port of merge_func_code functionality from probfit to iminuit

mbaak:
Since probfit is depracated, it would make sense (imho) to include this useful bit of code 
here, just like decribe() and make_func_code().
It makes it much easier to construct cost functions from compound objects.

I have added an example to:
generic_least_squares_function.ipynb

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,6 +5,16 @@
 Changelog
 =========
 
+2.4.1 (March 6, 2021)
+---------------------
+
+New features
+~~~~~~~~~~~~
+- ``iminuit.user.merge_user_func``, ported from probfit, by @mbaak
+  This functionality makes it possible to merge function codes into compound objects,
+  allowing one to build more complex cost functions easily.
+  See for example ``tutorials/generic_least_squares_function.ipynb``.
+
 2.4.0 (February 10, 2021)
 -------------------------
 
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -962,3 +962,113 @@ def _address_of_cfunc(fcn):
     if isinstance(fcn, c_sig):
         return cast(fcn, c_void_p).value
     return 0
+
+
+def merge_func_code(
+    *arg, prefix=None, skip_first=False, factor_list=None, skip_prefix=None
+):
+    """
+    Merge function arguments.
+
+        def f(x,y,z): return do_something(x,y,z)
+        def g(x,z,p): return do_something(x,y,z)
+        fc, pos = merge_func_code(f,g)
+        #fc is now ('x','y','z','p')
+
+    Parameters
+    ----------
+        - **prefix** optional array of prefix string to add to each function
+          argument. Default None.::
+
+                def f(x,y,z): return do_something(x,y,z)
+                def g(x,z,p): return do_something(x,y,z)
+                fc, pos = merge_func_code(f,g,prefix=['f','g'], skip_first=True)
+                #fc now ('x','f_y','f_z','f_p')
+
+        - **skip_first** option boolean to skip prefixing the first argument
+          of each function. This should normally be true when prefix is not
+          None.
+
+        - **factor_list** option list of functions. For function that skip_first
+          should not be applied. The choice of prefix applice to factor_list
+          will be in the same order as prefix. Default None
+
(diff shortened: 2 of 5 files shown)


mayou36:
So, as far as I understand, probfit was mainly deprecated due to [zfit](https://github.com/zfit/zfit) that is meant to me a more extensive library especially to build more complicated cost functions.

I tend to contribute it there, if the functionality is not yet provided there, instead of here to keep this primarily the repository for Minuit.

Feel free to open an issue in zfit if you think there should be changes in general, it aims to take care of all the cost function building, so let us know if it misses generality or similar.

mbaak:
zfit looks quite nice, I didn't know about this until only recently, but it seems very much written for continuous variables only, described with probability densities. I’ve not seen much (or any?) support for categorical features, which is what 
I need most in practice. 

I like iminuit and I'm happy to plug in my own cost function, and write probability mass functions on top of that. Functions like describe() and make_func_code() are quite useful for that, and I saw are already part of iminuit. (Previously part of probfit I believe.) I’m just making a suggestion to add a bit more user-functionality for that. 

Up to you of course. But please realize that not everyone is using zfit, there are many standalone use cases of iminuit as well, for which this should be useful.

mayou36:
> I’ve not seen much (or any?) support for categorical features, which is what
> I need most in practice.
This is a good point, while it can be used, it is indeed not optimized for this.
Out of curiosity (and feel free to answer by mail jonas.eschle at cern.ch), what are the things you often do? I would be quite interested to know in order to learn more usecases.

> I like iminuit and I'm happy to plug in my own cost function, and write probability mass functions on top of that. Functions like describe() and make_func_code() are quite useful for that, and I saw are already part of iminuit. (Previously part of probfit I believe.) I’m just making a suggestion to add a bit more user-functionality for that.

Fully agree, it is indeed nice to use it standalone. My concern is just not to overload iminuit, so if there would be a different place, e.g. a different repo in the ecosystem that fits better for the code, that could be preferable

mbaak:
(Ah, I see now you're one of the authors of zfit :))

I saw that iminuit already has some code to help out users, e.g. multiple cost functions and the describe() and make_func_code() functions needed to build those yourself. So I figured this would be a good place for the merge_func_code() function as well, as it's quite useful for users to build more complex cost functions. 

I understand your concerns - this is perhaps an edge case - but I'm not aware of a better repo for this than iminuit (clearly probfit is deprecated), and imho it fits well with the utility code that's already there.

I like iminuit, I know Minuit well from my background in hep, it's hard to beat as an optimizer.
These days I work a lot with with so-called "weak labels" and bayesian networks. (For weak labels see: https://arxiv.org/pdf/1810.02840.pdf) Basicly, I work with probabilistic models for categorical variables, 
ie. mass functions not densities. 

Anyway, my last try. :)

HDembinski:
@mbaak I think what you want to do here you can do better by inheriting from iminuit.cost.Cost. You can add cost functions to combine them.

HDembinski:
@mbaak Sorry for the slow reply, but I am totally overworked since about January and will continue to be overworked until the end of March. Afterwards I expected to have again more time to react here on Github. Please don't be disappointed that I didn't react sooner.

I hope I didn't misunderstand your patch, but I think users can do what you implement here already by inheriting from iminuit.cost.Cost. Could you check whether that fulfills your needs? I don't want to add two ways of doing the same thing to iminuit. The API with iminuit.cost.Cost is nicer, it is high-level.

mbaak:
Yes, I will have a good look at this code.

mbaak:
@HDembinski I've had a look at the Cost and CostSum code. I'm afraid I can't use that, at least not easily or without a rewrite. I can see how this code works well for simultaneous fits or probability-products. But as it happens as for my cost function I'm taking the sum of several probabilities, and then the log of that, and I can't split that into a sum of cost functions. (I.e. I need merge_func_code() to add the probability terms.) Perhaps you can see another way to reuse the code but I'm afraid I don't.

HDembinski:
Ok, I see. Yes, Cost is really for combining measurements and for combining a cost function and a penalty term. I understand now. I am not sure whether this is the ideal API to do this, but I don't have a better idea and I can see how it is useful. I will merge this.

PR #619, opened 2021-03-14
Fix: adjust limits in `draw_profile`

mayou36:
Plotting the error as an area can expand the plot if the requested span is smaller than the errors.

Fixed by respecting the axes limits.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -1180,7 +1180,7 @@ class Minuit:
 
         if vmin is not None and band:
             # make sure not to increase the plots size if errors > axes limits
-            # do not change vmin, vmax as title should contain the uncertainties, not plot limits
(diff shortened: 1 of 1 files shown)


HDembinski:
Thanks for the patch!

HDembinski:
I had to revert this. The span is supposed to represent the error. Your patch makes it potentionally shorter. This is not a "fix" this is a change in behavior.

HDembinski:
You can restrict the x-range again after calling draw_profile with plt.xlim without changing how the band is drawn.

mayou36:
I agree, indeed, I did not think far enough! It is maybe worth to mention that in the docs? I agree on the behavior, you are absolutely right on this, but plotting simply a profile within a range it is expected to be a plot in this range and not larger and may surprise users. Just as a remark maybe? But however you feel about it, if you would argue that it is actually obvious, I am also fine to make no remark.

Issue #272, opened 2018-07-07
Should values and errors be dict or view objects?


cdeil:
I'm trying out `iminuit` for the first time since your recent changes and noticed this:
```
In [1]:     from iminuit import Minuit
   ...: 
   ...:     def f(x, y, z):
   ...:         return (x - 2) ** 2 + (y - 3) ** 2 + (z - 4) ** 2
   ...: 
   ...:     m = Minuit(f)
   ...: 
   ...:     m.migrad()  # run optimiser
   ...:     print(m.values)  # {'x': 2,'y': 3,'z': 4}
   ...: 
   ...:     m.hesse()   # run covariance estimator
   ...:     print(m.errors)  # {'x': 1,'y': 1,'z': 1}
   ...: 
   ...: 
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter x does not have initial value. Assume 0.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter x is floating but does not have initial step size. Assume 1.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter y does not have initial value. Assume 0.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter y is floating but does not have initial step size. Assume 1.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter z does not have initial value. Assume 0.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: Parameter z is floating but does not have initial step size. Assume 1.
  
/Users/deil/software/anaconda3/envs/iminuit-dev/bin/ipython:6: InitialParamWarning: errordef is not given. Default to 1.
  
ValueView{ 'x'=2.0000000000047327,
 'y'=3.000000000007099,
 'z'=4.000000000009465
}
ErrorView{ 'x'=1.0000000000000402,
 'y'=0.9999999999999426,
 'z'=1.0000000000001645
}

In [2]: m.errors
Out[2]: <iminuit._libiminuit.ErrorView at 0x11a1216a0>

In [3]: m.errors[0]
Out[3]: 1.0000000000000402
In [8]: json.dumps(m.errors)
---------------------------------------------------------------------------
[...]

PR #620, opened 2021-03-18
Add to_dict method to BasicView

watsonjj:
The new version of ValueView's `__iter__` now only iterates over the values, instead of keys & values. This is nice for converting to numpy arrays but removes the ability to convert to a dict with `dict(m0.values)`. This PR adds a `to_dict` method to `BasicView` to conveniently convert views to a dict.

Relevant historic discussion: https://github.com/scikit-hep/iminuit/issues/272#issuecomment-403433623

### DIFF
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -79,6 +79,10 @@ class BasicView:
         s += ">"
         return s
 
+    def to_dict(self):
+        """Obtain dict representation"""
+        return dict(zip(self._minuit._pos2var, self))
+
 
 def _ndim(obj):
     nd = 0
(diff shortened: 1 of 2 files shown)


HDembinski:
Thanks for the patch. What is the use case for this?

watsonjj:
I often convert the `ValuesView` to a built-in object types to interface it with other packages. And I prefer a dict as I do not lose reference of which value corresponds to which parameter. One example is storage into a Pandas Dataframe (each row is a separate dataset, each column is a parameter from the fit).

HDembinski:
Ok, thanks for the explanation. The interface also looks pandas-inspired, which is good.

Issue #621, opened 2021-03-18
Improved support for simultaneous fitting


watsonjj:
A nice feature of probfit is the `SimultaneousFit` class (https://probfit.readthedocs.io/en/latest/api.html#simultaneous-fit). 

It appears that iminuit itself is already mostly capable of the features that `SimultaneousFit` provides (through its `CostSum` class). However, unless I am mistaken, there are a few convenience methods missing that would complete this feature. As probfit is deprecated, I would like to contribute this additional functionality to iminuit.

My motivation for this is my single-photoelectron spectra fitting package (https://gitlab.cta-observatory.org/cta-consortium/aswg/tools/spefit), which already utilises iminuit. I was previously unaware of these existing simultaneous-fit features when originally writing this package. I would like to release a new version of `spefit` that replaces much of its own simultaneous-fit bookkeeping with that already provided by iminuit.

Here I demonstrate a version of the probfit `SimultaneousFit` example (https://probfit.readthedocs.io/en/latest/pyplots/costfunc/simul.py) which solely utilises iminuit to achieve the same result:

```python
from iminuit import Minuit
from iminuit.cost import UnbinnedNLL
from iminuit.util import make_func_code
from matplotlib import pyplot as plt
import numpy as np
from numpy.random import randn, seed
from numba import vectorize, float64
from math import exp, sqrt, pi

SQRT2PI = sqrt(2.0 * pi)

@vectorize([float64(float64, float64, float64)], fastmath=True)
def normal_pdf(x, mean, std_deviation):
    u = (x - mean) / std_deviation
    return exp(-0.5 * u ** 2) / (SQRT2PI * std_deviation)

seed(0)
width = 2.0
data1 = randn(1000) * width + 1
data2 = randn(1000) * width + 2

normal_pdf.func_code = make_func_code(("x", "mu_1", "sigma"))
lh1 = UnbinnedNLL(data1, normal_pdf)
normal_pdf.func_code = make_func_code(("x", "mu_2", "sigma")) # TODO: Add method to rename lh1
lh2 = UnbinnedNLL(data2, normal_pdf)

simlh = lh1 + lh2

m = Minuit(simlh, mu_1=1.2, mu_2=2.2, sigma=1.5)

def plot(cost, minuit, i, ax):
    names_i = [minuit.parameters[map_i] for map_i in simlh._maps[i]]  # TODO: Add method to BasicView to obtain view subset
[...]

PR #624, opened 2021-03-22
Improved support for simultaneous fitting

watsonjj:
As outlined in #621, I believe there are some minor changes to the Cost classes that would improve support for simultaneous fitting in iminuit. This PR includes the following changes:
- Setter for a Cost function's func_code
- Method to rename a single parameter in the model of a Cost function (`rename_parameter`)
- Enable public access to the CostSum._maps parameter
- Tutorial demonstrating simultaneous fitting of multiple datasets which share some model parameters


HDembinski:
I have problems with this API. You should not change the func_code of the Cost functions. You should change the func_code of the model.

HDembinski:
Please check https://github.com/scikit-hep/iminuit/issues/625 for an alternative interface.

HDembinski (changes requested):
I think this is mostly superseded now with the other pending changes. This PR should only handle `maps`, I think.

Issue #625, opened 2021-03-23
BasicView should accept list of keys and return a sequence of the values


HDembinski:
@watsonjj wants in https://github.com/scikit-hep/iminuit/issues/621 to have easy access to the parameter names and fitted values for a cost function that is part of the a CostSum. We should mimic the numpy array interface for this, which allows one to pass a list of indices to get a numpy array of the values at those key positions.

```py
m = iminuit.Minuit(..., mu=1, sigma=2, beta=3)

m.values[["mu", "beta"]] # returns [1, 3]
m.values[["beta", "sigma"]] # returns [3, 2]
# etc

# using indices should also work
m.values[[0, 2]] # returns [1, 3]
m.values[[2, 1]] # returns [3, 2]

# likewise for m.parameters, m.covariance, m.fixed
```

HDembinski:
@watsonjj This together with some way to get the parameter names per Cost function in a CostSum would work for you, no?

watsonjj:
Indeed.

watsonjj:
In my experimenting for #621 I did implement this for ValueView. However I wasn't sure on how to do the same for the parameter names (I had to stick with a list comprehension). I can open a PR with just my implementation for the ValueView access.

watsonjj:
The reason I included the ability to use `__getitem__` with a tuple in #627 was because `CostSum._maps[i]` returns tuples. Do we intend the user to have to convert that tuple into a list before passing to `ValueView.__getitem__`?

watsonjj:
The ability to get the parameter names in this same way is still missing. Currently have to do `[minuit.parameters[p] for p in cost._maps[i]]`

HDembinski:
> The reason I included the ability to use `__getitem__` with a tuple in #627 was because `CostSum._maps[i]` returns tuples. Do we intend the user to have to convert that tuple into a list before passing to `ValueView.__getitem__`?

Not really, I would like to allow any sequence, but I was worried that this may cause trouble. If you have a 2D numpy array called `arr`, then `arr[(1, 2)]` and `arr[[1, 2]]` do very different things. I am trying to mimic the numpy array API here, that's why I am concerned about allowing tuples.

HDembinski:
For 1D arrays implemented with BasicView allowing tuples is safe, I think, but people will expect that the same syntax works for util.Matrix, which derives from numpy.ndarray and there it will fail if you use tuples.

PR #627, opened 2021-03-23
Enable BasicView __getitem__ with tuple/list

watsonjj:
As described in #625, this PR seeks to add the ability to access a subset of the ValueView (and other views) using a list/tuple of names/indices.

### DIFF
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -45,9 +45,13 @@ class BasicView:
 
     def __getitem__(self, key):
         """
-        Get subset of view.
+        Get values of the view.
 
-        Key can be an index,a parameter name, a tuple/list, or a slice.
+        Parameters
+        ----------
+        key: int, str, slice, list of int or str
+              If the key is an int or str, return corresponding value.
+              If it is a slice, list of int or str, return the corresponding subset. 
         """
         key = _key2index(self._minuit._var2pos, key)
         if isinstance(key, slice):
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -44,11 +44,17 @@ class BasicView:
         return self._minuit.npar
 
     def __getitem__(self, key):
-        """Get value at key, which can be an index, a parameter name, or a slice."""
+        """
+        Get subset of view.
+
+        Key can be an index,a parameter name, a tuple/list, or a slice.
+        """
         key = _key2index(self._minuit._var2pos, key)
         if isinstance(key, slice):
             ind = range(*key.indices(len(self)))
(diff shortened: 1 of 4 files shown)


src/iminuit/util.py
```diff
@@ -44,11 +44,17 @@ def __len__(self):
         return self._minuit.npar
 
     def __getitem__(self, key):
-        """Get value at key, which can be an index, a parameter name, or a slice."""
+        """
+        Get subset of view.
+
+        Key can be an index,a parameter name, a tuple/list, or a slice.
+        """
```
HDembinski, line 51:
```suggestion
        """
        Get values of the view.

        Parameters
        ----------
        key: int, str, slice, list of int or str
              If the key is an int or str, return corresponding value.
              If it is a slice, list of int or str, return the corresponding subset. 
        """
```

src/iminuit/util.py
```diff
@@ -44,11 +44,17 @@ def __len__(self):
...
         key = _key2index(self._minuit._var2pos, key)
         if isinstance(key, slice):
             ind = range(*key.indices(len(self)))
             return [self._get(i) for i in ind]
+        elif isinstance(key, list):
+            return [self._get(i) for i in key]
```
HDembinski, line 57: This does not work for list of strings.

HDembinski: The functionality should be implemented in _key2index.

src/iminuit/util.py
```diff
@@ -947,6 +953,8 @@ def _key2index(var2pos, key):
         start = _key2index(var2pos, sl.start) if sl.start is not None else None
         stop = _key2index(var2pos, sl.stop) if sl.stop is not None else None
         return slice(start, stop, sl.step)
+    if isinstance(key, tuple) or isinstance(key, list):
```
HDembinski, line 956: Only lists are allowed here to be one the safe side. Tuples have special meaning in getitem/setitem.

tests/test_util.py
```diff
@@ -77,6 +77,11 @@ def test_ValueView():
     assert v_dict["y"] == v["y"]
     assert v_dict["z"] == v["z"]
 
+    assert_equal(v[["x", "z", "y"]], (2, 3, 2))
+    assert_equal(v[("x", "z", "y")], (2, 3, 2))
```
HDembinski, line 81: Only lists are allowed.

HDembinski (changes requested):

tests/test_util.py
HDembinski: The tests are incomplete, you also need to test assignment.

PR #628, opened 2021-03-23
replace merge_func_code with merge_signatures, add make_func

### LINKED ISSUE #623: Simplify merge_func_code
@mbaak I merged your patch but I am not at all happy with this function. I am considering to revert this at least partially.

Is this interface needed? What is the use case for each option?
```
        - **skip_first** option boolean to skip prefixing the first argument
          of each function. This should normally be true when prefix is not
          None.

        - **factor_list** option list of functions. For function that skip_first
          should not be applied. The choice of prefix applice to factor_list
          will be in the same order as prefix. Default None

        - **skip_prefix** list of variable names that prefix should not be
          applied to.
```
factor_list should be handled by skip_first. skip_first can either be a boolean or a sequence of boolean with the same length as the list of functions to be merged. The boolean for each function then tells you to skip or not. skip_prefix should be removed.

### LINKED ISSUE #626: API to rename parameters
It seems there is a need to rename parameters in some functions, especially when combining likelihoods (what @watsonjj calls simultaneous fits).

It is already possible to rename parameters like this:

```py
def normal_pdf(x, mu, sigma):
   # ...

def my_normal_pdf(x, mu_1, sigma_1):
      return normal_pdf(x, mu_1, sigma_1)
```

For common use this is a bit too much to type. It was proposed in #621 to handle this in the Cost function, but that is not the right place for such a functionality, since it is quite general and it should be applied to the model and not the cost.

We want something like this:
```py
# replace all parameter names
f = make_with_signature(normal_pdf, "x", "mu_1", "sigma_1")
g = make_with_signature(normal_pdf, "x", "mu_2", "sigma_2")

# renaming only some parameters
h = make_with_signature(normal_pdf, mu="mu_1") 
```

@watsonjj and @mbaak please comment.

HDembinski:
Closes #626
Closes #623

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@
 repos:
 # Standard hooks
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v3.2.0
+  rev: v3.4.0
   hooks:
   - id: check-case-conflict
   - id: check-docstring-first
@@ -42,7 +42,7 @@ repos:
 
 # Python linter (Flake8)
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.8.4
+  rev: 3.9.0
   hooks:
   - id: flake8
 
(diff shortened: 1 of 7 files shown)


HDembinski:
@mbaak This adds `merge_signature` which replaces merge_func_code. It does not return a func_code object, just the joined arguments as a tuple and the mapping. This is a low-level utility. To get the same as merge_func_code, you need to call make_func_code separately on the first argument. The reason to not combine these two operations in one function is to have a minimal set of compose-able functions. I use merge_signature in CostSum now, where it replaces the original implementation.

@mbaak @watsonjj To rename parameters of a function, I added `make_func` which is another name for `make_with_signature` in #626

HDembinski:
> I am open to adding another high-level function, perhaps called merge_func, that returns new functions that accept the merged argument tuple.

I added such a function called `expanded` to the new module `experimental.py`.

PR #629, opened 2021-03-24
Create CostSum.signature_map

watsonjj:
Enable public access to the CostSum._maps parameter, taking into account the comments in #624

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -273,6 +273,36 @@ class CostSum(Cost, Sequence):
         """Get constituent cost function by index."""
         return self._items.__getitem__(key)
 
+    @property
+    def signature_map(self):
+        """
+        Access the signature map for the underlying Cost functions.
+
+        The values in the signature map represent the position of a parameter
+        in this CostSum's signature::
+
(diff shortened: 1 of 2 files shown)


HDembinski:
@watsonjj Thank you for another quick patch, but we should really discuss this a bit before rushing to an implementation.

We don't want needless redundancy in the iminuit API, so the question is whether you can get the effect already with the API that we offer.

```py
from iminuit.cost import LeastSquares
from iminuit.util import describe
from iminuit import Minuit

def model1(x, a):
    return x + a

def model2(x, a, b):
    return x + a + b

ls1 = LeastSquares([1], [2], [3], model1)
ls2 = LeastSquares([1], [2], [3], model2)
ls = ls1 + ls2
m = Minuit(ls, a=1, b=2)
m.values[list(describe(ls[0])] # returns [1]
m.values[list(describe(ls[1])] # returns [1, 2]
```

watsonjj:
Thanks! Now I see there was already a way to achieve what I needed :)

Yeah, being allowed to pass tuples into `__getitem__` would be cleaner for my use-case

PR #634, opened 2021-03-26
describe returns list; util.Matrix supports special selection

### LINKED ISSUE #625: BasicView should accept list of keys and return a sequence of the values

HDembinski:
Closes #625 at the cost of changing the output type of describe back from tuple to list

This is patching two things. Firstly, `describe` is now again returning a list (as it used to in v1.x), which works better with the extended indexing capabilities of ValueView and Matrix. Secondly, Matrix now handles slices and selections in a special way so that you get the right submatrix. Example code:

```py
c1 = LeastSquares(...) # parameters a, b
c2 = LeastSquares(...) # parameter b, c

m = Minuit(c1 + c2, ...)

# select only a,b
c1args = describe(c1)
value_only_ab = m.values[c1args]
cov_only_ab = m.covariance[c1args] # this now returns the correct square matrix
```

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,23 +5,31 @@
 Changelog
 =========
 
-2.4.1 (March 6, 2021)
----------------------
+2.5.0
+-----
 
 New features
 ~~~~~~~~~~~~
-- ``iminuit.user.merge_user_func``, ported from probfit, by @mbaak
-  This functionality makes it possible to merge function codes into compound objects,
-  allowing one to build more complex cost functions easily.
-  See for example ``tutorials/generic_least_squares_function.ipynb``.
-- ``iminuit.util.BasicView.to_dict`` added, by @watsonjj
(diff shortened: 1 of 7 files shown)


HDembinski:
@watsonjj @mbaak Any comments?

mbaak:
@HDembinski Looks good to me!

PR #636, opened 2021-03-31
fixed missing ',' in iminuit BibTeX.

The-Ludwig:
Just saw that one ',' is missing in the BibTeX example entry, after trying to copy it.


HDembinski:
Thanks!

PR #657, opened 2021-05-19
chore: pypa/cibuildwheel

henryiii:
Updating for the new location as part of the Python Packaging Authority!


HDembinski:
Thanks!

PR #659, opened 2021-05-25
Add scipy minimizers

### LINKED ISSUE #654: Allow calling scipy and nlopt minimizers from Minuit object
We already support mixing Minuit with other minimisers, but we do not provide a convenient interface for this. Something like this should work:
```py
def fcn(a, b, c): [...]

m = Minuit(fcn, a=1, b=2, c=3)
m.scipy(method=...) # only works if scipy is installed
# optionally run m.migrad()
m.hesse()

# OR

m = Minuit(fcn, a=1, b=2, c=3)
m.nlopt(method=...) # only works if nlopt is installed
m.hesse()
```
For this to work, scipy and nlopt have to get access to the parameter limits and only see the parameters that are not fixed. The results need to be converted into Minuits internal C++ structures.

HDembinski:
Partially closes #654, a separate issue should be opened for supporting nlopt and some Bayesian optimization algorithm (e.g. hyperopt).

- [x] Adds the ability to construct a FunctionMinimum from scratch with properly calculated edm based on external estimates of the gradient and inverse of the Hessian matrix.
- [x] Fixes a bug when MnHesse is used to add a covariance matrix to an existing FunctionMinimum, which would set the status of the FunctionMinimum to valid even if it was invalid before.
- [x] Param.has_limits, Param.has_lower_limit, Param.has_upper_limit are now properties; a missing limit is indicated by setting the value to `None`.
- [x] The uninformative `Valid Parameters` status was removed from the text/html representations of FMin. A FunctionMinimum returned from iminuit always has "Valid Parameters". It matters that the function minimum is valid or not.
- [x] A title was added to the text/html representations of FMin to indicate which minimization algorithm was used.
- [x] Lots of added static type info.
- [x] BasicView is now a proper abstract base class.
- [x] Tutorial
- [x] Changelog entry
- [x] support the same calling convention for the scipy constraints that the main function uses (constraints should use parameters a, b, c if the main function uses a, b, c)


HDembinski:
@jpivarski @matthewfeickert Requesting comments on this PR if you have any.

The integration of scipy minimizers into iminuit is a big step, because when this works technically with scipy, I can also integrate other interesting optimization methods with the same underlying machinery. Two commonly requested features by users are global optimization and constrained optimization. The latter is more powerful than placing box bounds on parameters and would be very useful for HEP and provided by scipy. To support global optimization, I am considering to add Bayesian optimization, probably based on hyperopt. The patch is pretty large, sorry about that. I think you don't need to check the implementation details (although I am happy for feedback)

jpivarski (approved):
I scanned all the differences (it's a large PR) and nothing jumped out as looking like a bad choice. I'm impressed by the thoroughness of the typing.

As a more big-picture question, what's the motivation for SciPy minimizers with an iminuit interface, rather than iminuit with a SciPy interface? (Maybe the latter already exists and I don't know about it.) I would have thought that SciPy would be the standard that interfaces have been designed around, though I'm sure it's bare-bones compared to all of the interactive features in iminuit (e.g. the HTML reports and Jupyter integration). Is that it? Is it so that any minimizer with a SciPy interface can now be used in this interactive-friendly way?

HDembinski:
@jpivarski You can already use Minuit with the scipy minimize interface, but that is not really all that useful, because a central idea behind Minuit is the ability to do a complex minimization in several steps, by using a combination of different methods and by fixing/releasing paramters, setting/removing bounds. Of course this can also be achieved in different ways, but it is particularly easy with the Minuit interface to do multistage fits. This adds a bunch of scipy minimizers to iminuits toolbox and it helps to promote their special features to HEP.

So all in all the answer is: you could achieve what this does before, but this patch is making it easier.

matthewfeickert:
Whoa, sorry @HDembinski. I think I missed the original notification on this &mdash; didn't mean to ignore this! I'll try to get to it later tonight, but if you wanted to merge before then please don't wait for me.

HDembinski:
@matthewfeickert No problem if you are busy, I think I have figured everything out. There are some caveats to wrapping scipy inside Minuit that cannot be fully resolved, but I am confident about my solutions to these caveats.

matthewfeickert:
As a note to future @matthewfeickert (not to Hans &mdash; you're fine), this is causing the `pyhf` nightly [HEAD of dependencies CI to fail](https://github.com/scikit-hep/pyhf/runs/2750920068?check_suite_focus=true) on the [`32b` `test_optim.py` tests](https://github.com/scikit-hep/pyhf/blob/03b914b1aa9006d8aa8ca5ecbbd2e6b6e6bdd1ce/tests/test_optim.py#L31-L48) which we know from previous experience are probably not great choices. The `pyhf` devs should fix this probably at the same time they fix https://github.com/scikit-hep/pyhf/pull/1057.

cc @kratsg

PR #670, opened 2021-07-17
chore: bump to cibuildwheel 2

henryiii:


HDembinski:
@henryiii Thanks, I suppose this means that https://github.com/scikit-hep/iminuit/blob/develop/.github/workflows/wheels.yml#L6-L9 can be removed?

Issue #676, opened 2021-08-03
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, pip install iminuit builds the wheels from source code and then install it. It requires user to have development environment installed on his system. also, it take some time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, pip install iminuit should download the wheels from pypi

@HDembinski, Please let me know your interest on releasing aarch64 wheels. I can help in this.

HDembinski:
Thank you for the contribution. Building the wheels now takes forever, but I guess we can live with that.

PR #677, opened 2021-08-03
Add aarch64 test and  wheel build support

odidev:
Added aarch64 test and  wheel build support.
Related to #676 . @HDembinski, Could you please review this PR?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,29 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: python -m pip install --upgrade pip wheel
-    - run: python -m pip install .
-    - run: python -m pip install pytest
+    - run: python -m pip install .[test]
     - run: python -m pytest
+
+  test-aarch64:
+    strategy:
+      matrix:
+        pyver: [cp39-cp39]
+        arch: [aarch64]
+      fail-fast: false
+    runs-on: ubuntu-latest
+    env:
+      py: /opt/python/${{ matrix.pyver }}/bin/python
+      img: quay.io/pypa/manylinux2014_${{ matrix.arch }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - uses: docker/setup-qemu-action@v1
+    - run: >
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
+            ${{ env.img }}
+            bash -exc '${{ env.py }} -m venv venv &&
+            source venv/bin/activate &&
+            python -m pip install --upgrade pip setuptools wheel &&
+            python -m pip install .[test] &&
+            python -m pytest'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: python -m pip install --upgrade pip wheel
-    - run: python -m pip install .[test]
+    - run: python -m pip install . pytest
     - run: python -m pytest
 
   test-aarch64:
@@ -52,6 +52,6 @@ jobs:
             ${{ env.img }}
             bash -exc '${{ env.py }} -m venv venv &&
             source venv/bin/activate &&
-            python -m pip install --upgrade pip setuptools wheel &&
-            python -m pip install .[test] &&
+            python -m pip install --upgrade pip wheel &&
+            python -m pip install . pytest &&
             python -m pytest'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,26 +32,29 @@ jobs:
     - run: python -m pip install . pytest
     - run: python -m pytest
 
-  test-aarch64:
-    strategy:
-      matrix:
-        pyver: [cp39-cp39]
-        arch: [aarch64]
-      fail-fast: false
-    runs-on: ubuntu-latest
-    env:
-      py: /opt/python/${{ matrix.pyver }}/bin/python
-      img: quay.io/pypa/manylinux2014_${{ matrix.arch }}
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        submodules: true
-    - uses: docker/setup-qemu-action@v1
-    - run: >
-            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
-            ${{ env.img }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+      uses: docker/setup-qemu-action@v1
+    - name: Install tools
+      run: |
+        docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+          ${{ env.img }} ${{ env.py }} -m pip install -U  pip build
+    - name: Install dependencies
+      run: |
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '${{ env.py }} -m pip install virtualenv && ${{ env.py }} -m venv .env && \
```
HDembinski, line 62: Is pip install virtualenv necessary? venv is part of pythpn since 3.3.

HDembinski (commented):
Looks largely ok to me, let's see what Henry has to say then this is good to go. Thanks for the contribution!

.github/workflows/test.yml
odidev: Thanks. Its not required. Removed it.

.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+    - name: Install tools
+      run: |
+        docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+          ${{ env.img }} ${{ env.py }} -m pip install -U  pip build
```
henryiii, line 57: What is this step for? You are not using `build` later, and `venv` installs `pip` from the built-in copy (which means you need to up date it after installing, but you do that correctly below). Can you just remove this step?

.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+    - name: Install dependencies
+      run: |
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '${{ env.py }} -m venv .env && \
+            source .env/bin/activate && \
+            python -m pip install --upgrade pip setuptools wheel && \
+            deactivate'
```
henryiii, line 65: Stylistic choice (so it's up to @HDembinski), but the `/`'s make this harder to read and are easier to forget when modifying this, IMO:

```suggestion
      run: >-
            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
            ${{ env.img }}
            bash -exc '${{ env.py }} -m venv .env &&
            source .env/bin/activate &&
            python -m pip install --upgrade pip setuptools wheel &&
            deactivate'
```

.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+    - name: Test with pytest
+      run: |
+          docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '\
+            source .env/bin/activate && \
+            python -m pip install . && \
+            pip install pytest && \
+            python -m pytest && \
+            deactivate'
```
henryiii, line 75:
```suggestion
      run: >-
          docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
            ${{ env.img }}
            bash -exc '
              source .env/bin/activate &&
              python -m pip install . &&
              pip install pytest &&
              python -m pytest &&
              deactivate'
```
Even more so here, where they hide the quotes being used on the command, IMO.

henryiii: Isn't there a testing extra? If that's the case, you should do `python -m pip install .[test]`, rather than manually installing pytest. If we wanted to add a pytest plugin in the future, for example, or restrict pytest, etc, it's better to use a centrally managed extra.

Also, you don't need to activate and deactivate the environment, you can just call `.env/bin/python -m ...`

Please name virtual environments `.venv` instead of `.env`. That's actually becoming a standard name; the Python launcher for Unix even supports it directly.

henryiii: Also, PEP 518 means you don't have to install/upgrade `setuptools` & `wheel` anymore, they are specified in pyproject.toml.

henryiii (approved):
Thanks for working on this! I have had "add Linux ARM and macOS ARM" on my extended todo list for a while. I have some minor cleanup and stylistic suggestions, but overall this looks good to me. Now I only have macOS ARM. :)

.github/workflows/test.yml
HDembinski: I agree with Henry :)

.github/workflows/test.yml
henryiii:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+        docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+          ${{ env.img }} ${{ env.py }} -m pip install -U  pip build
+    - name: Install dependencies
+      run: >
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
+            ${{ env.img }}
+            bash -exc '${{ env.py }} -m venv .env &&
+            source .env/bin/activate &&
+            python -m pip install --upgrade pip setuptools wheel &&
+            deactivate'
```
henryiii, line 65:
```suggestion
            bash -exc '${{ env.py }} -m venv .env &&
              .env/bin/python -m pip install --upgrade pip setuptools wheel'
```

.github/workflows/test.yml
```diff
@@ -32,3 +32,44 @@ jobs:
...
+    - name: Test with pytest
+      run: >
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws
+            ${{ env.img }}
+            bash -exc '
+              source .env/bin/activate &&
+              python -m pip install . &&
+              pip install pytest &&
+              python -m pytest &&
+              deactivate'
```
henryiii, line 75:
```suggestion
              .venv/bin/python -m pip install .  pytest &&
              .venv/bin/python -m pytest'
```

PR #678, opened 2021-08-03
chore: add Apple Silicon wheels

henryiii:
- chore: sync with cmake_example

I've made this look a bit more like `pybind/cmake_example` (and made `cmake_example` look a bit more like this in https://github.com/pybind/cmake_example/pull/53, as well) - the more this looks like the maintained example, the better. I've tried not to lose anything that was was an intentional change from that example, like the usage of ninja on Windows, etc. I've also kept f-strings and Paths, anything that was clearly an improvement over the original (and I think we are likely about to remove Python 2/3.5 from our examples, so likely it will look even more similar in the future).

@HDembenski, if you don't like any changes, I can roll back any of them expect the key change that allows Apple Silicon to be supported. :)

- ci: build Apple Silicon wheels

I've added a universal2 wheel build for the only Pythons supported on Apple Silicon; 3.8 and 3.9. I've also removed "NumPy" from the build requirements - I've checked, and you never include "numpy.h" (from NumPy), but only `pybind11/numpy.h`, which does not require NumPy during the build.

### DIFF
--- a/cmake_ext.py
+++ b/cmake_ext.py
@@ -1,9 +1,20 @@
-import sys
 import os
+import re
 import subprocess
+import sys
+from pathlib import Path
+
 from setuptools import Extension
 from setuptools.command.build_ext import build_ext
-from pathlib import Path
+
+
+# Convert distutils Windows platform specifiers to CMake -A arguments
+PLAT_TO_CMAKE = {
+    "win32": "Win32",
+    "win-amd64": "x64",
+    "win-arm32": "ARM",
+    "win-arm64": "ARM64",
+}
 
 
 class CMakeExtension(Extension):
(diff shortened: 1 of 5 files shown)


henryiii:
```
AssertionError: pypi version exists
```

Guessing this means the versions needs to be changed to a dev version or something like that? This produces a lot more AssertionErrors locally.

HDembinski:
Yes, you need to increment the version number in iminuit if the current version is still the latest release. And you need to add a new changelog entry for that version.

HDembinski:
@henryiii I don't like the global PLAT_TO_CMAKE constant. This is removing the lookup table from where it is used and makes the code harder to read. The lookup appears in exactly one place and the table is not very long, so there is no reason to move it out of context.

HDembinski:
Now it is good to go, thank you for the contribution.

henryiii:
FYI, Python 3.10 rc 1 is out, and filtering down the chain - if you'd like to produce 3.10 wheels with RC 1, it should be possible in a day or two.

cmake_ext.py
```diff
@@ -56,7 +56,7 @@ def build_extension(self, ext):
             # Cross-compile support for macOS - respect ARCHFLAGS if set
             archs = re.findall(r"-arch (\S+)", os.environ.get("ARCHFLAGS", ""))
             if archs:
-                cmake_args += ["-DCMAKE_OSX_ARCHITECTURES={}".format(";".join(archs))]
+                cmake_args += [f"-DCMAKE_OSX_ARCHITECTURES={';'.join(archs)}"]
```
henryiii: I didn't notice this, nice. :)

PR #679, opened 2021-08-05
Fix build error on Python 3.10 for Windows

cgohlke:
Since [bpo-11717: fix ssize_t redefinition error when targeting Win32 app](https://github.com/python/cpython/pull/24479), the Python headers no longer define the `ssize_t` type on Windows. Instead, use `Py_ssize_t`


HDembinski:
Thanks for this patch! Seems like a no-brainer to add, let me just see whether the tests all run through.

HDembinski:
@cgohlke Turns out I cannot reproduce the bug. Since 3.10 is not even released yet, I used 3.10.0-rc1 on Windows. Do you know a way to expose the bug in a test?

I don't want to use your change verbatim, because idiomatic C++ way to handle this is to use `pybind11::ssize_t` instead of the C-ish macro definition Py_ssize_t. In `type_caster.hpp`, it should not be necessary to replace ssize_t, because that is defined inside the pybind11 namespace where ssize_t should be found.

HDembinski:
@cgohlke Btw, I am curious why you want to use iminuit with an unreleased Python-3.10.

PR #683, opened 2021-08-23
fix: support cross-compiling with Conda

henryiii:
Credit to @tdegeus, see https://github.com/pybind/cmake_example/pull/58 and https://github.com/conda-forge/iminuit-feedstock/pull/62.

### DIFF
--- a/cmake_ext.py
+++ b/cmake_ext.py
@@ -31,6 +31,11 @@ class CMakeBuild(build_ext):
             f"-DCMAKE_BUILD_TYPE={cfg}",
         ]
 
+        # Adding CMake arguments set as environment variable
+        # (needed e.g. to build for ARM OSx on conda-forge)
+        if "CMAKE_ARGS" in os.environ:
+            cmake_args += [item for item in os.environ["CMAKE_ARGS"].split(" ") if item]
+
         build_args = ["--config", cfg]  # needed by some generators, e.g. on Windows
 
         if self.compiler.compiler_type == "msvc":
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,6 +5,13 @@
 Changelog
 =========
 
+2.8.3
+-----
+
+Fixes
+-----
+- Support cross-compiling of ARM on Conda, by @henryiii
+
 2.8.2 (August 15, 2021)
 -----------------------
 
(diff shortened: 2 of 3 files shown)


HDembinski:
You need to increment the version number.

henryiii:
I know some projects do that, and that's fine, but personally, I don't like the practice of bumping the version in a PR. Three reasons I can think of: first, contributors don't know your version plans, so it can never be done in advance. Second, it tends to push toward one-release-per-PR, which is not ideal, patches often need to sit in the repo for a bit for stability. Finally, and most important, the patch generated by adding ".patch" to the end of the URL will now be broken - if I want to apply this patch to the existing release, I can't, because the patch will now contain a new version number change too, rather than being just this one fix. Conda-forge's patching system takes in-recipe patches, so I can manually edit the patch. Some build systems take the patch URL, which would not work. (In this case, the patch is already pulled out and ready to go, so no need to get the patch in. Conda-forge also is not picky about patching being in the upstream master before merging, like for example homebrew is.)

Anyway, totally fine to do this any way you want, just stating my general options not related to this PR, just the idea in general. Several packages do use this method, including a couple of other ones in Scikit-HEP.

henryiii:
There's also no need to make a release for conda, the patch is fine there. I don't think this will affect many users, as you have to be setting CMAKE_ARGS to run into it, which is mostly conda.

HDembinski:
I see all your points, this situation is not ideal, but I haven't been able to come up with something better yet.

The test makes sure that the version number is higher than the previously released version. There is no need to increase in the version number for every PR, only one increment between releases. Since this is the first PR since the last release, the test fails. It is not ideal, but it makes sure that I do not forget to increment the version number.

Another option that I can think of:
Restrict the version test to the master branch (would require some hacking). The releases are done from master, so pre-commit should detect the problem before I do the release. In worst case, if pre-commit does not catch this, the release job simply fails.

henryiii:
Ah, you are doing this the other way around - forcing a bump as soon as a release is made. That's better; I'd recommend you immediately do a bump push or PR after releasing, [like we do in pybind11](https://github.com/pybind/pybind11/blob/master/docs/release.rst#to-release-a-new-version-of-pybind11), for example. An automatic PR after release would be even nicer, I'll poke you if I ever do that for pybind11 or other project. The "master only" checking would probably be fine too, but not sure how easy that is to enable; probably possible, maybe within the check itself.

One downside with this is it's hard to guess what to bump right after a release; if you have a backport branch for a major/minor release, they need different bumps.

HDembinski:
The more I think about it, the more I believe that I just need to restrict the version check to the job that does the release. Before building all the wheels, I just need to check the version and number and fail early when the release manager forgot to increment it.

PR #688, opened 2021-09-28
chore: use PyPI package for clang-format

henryiii:
The pip-installable PyPI package is 1-2 MB, and has cibuildwheel wheels for all major platforms. It allows exact pinning of pre-commit versions rather than depending on what you have installed locally, and works on pre-commit.ci, allowing automatic fixes. It also comes with a pre-commit hook.  See https://github.com/ssciwr/clang-format-wheel or https://twitter.com/HenrySchreiner3/status/1442877941910802433?s=20 for my announcement. Versions 10, 11, and 12 are available.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,8 +51,8 @@ repos:
     files: src/iminuit/[^_].*\.py
 
 # C++ formatting
-- repo: git://github.com/doublify/pre-commit-clang-format
-  rev: 62302476d0da01515660132d76902359bed0f782
+- repo: https://github.com/ssciwr/clang-format-precommit
+  rev: v12.0.1
   hooks:
   - id: clang-format
 


henryiii:
PS: The pre-commit action is deprecated in favor of pre-commit.ci, so we could enable that now that all checks are "nice". It also gives you the optional auto fixes and the optional regular (weekly, monthly, or quarterly) update PRs.

henryiii:
```
TypeError: polyval requires ndarray or scalar arguments, got <class 'tuple'> at position 0.
```

Pretty sure I didn't break that. :)

HDembinski:
> PS: The pre-commit action is deprecated in favor of pre-commit.ci, so we could enable that now that all checks are "nice". It also gives you the optional auto fixes and the optional regular (weekly, monthly, or quarterly) update PRs. It's also _way_ faster.
> 
> Edit: I see you are not using the action, which makes it even slower still, as you loose all caching.

I had looked into using pre-commit.ci, but at the time there was a reason why I could not use it for iminuit. I use it for other projects. I don't remember the reason and whether it is still valid.

HDembinski:
@henryiii Please do not use tags like "chore:" for your commits, thanks.

henryiii:
I use the https://www.conventionalcommits.org convention unless told otherwise (and occasionally slip back even when told otherwise, so forgive me if I do and remind me again). Happy to follow whatever convention you like. I like conventionalcommits because it tells you where a commit would be listed in a changelog and what sort of version bump a commit requires. A feat: should not be in a patch release, a fix: probably should be backported (and ci:!), a chore: doesn't matter, and style can be skipped when blaming.

henryiii:
> but at the time there was a reason why I could not use it for iminuit.

Using local clang-format (or docker clang-format, or even conda clang-format until a few days ago) would be a reason. The slimmed down conda one (worked several days to get it under 250 MB) or the pip one (1-2MB) work fine on pre-commit.ci.

PR #693, opened 2021-10-13
rename clang-format hook to official mirror

henryiii:
I contributed types_or support to pre-commit's mirror tool, and now pre-commit mirrors clang-format directly, which removes the need for manual hook maintenance, so we are moving to using the official mirror. We are winding down the manual repository and suggesting users move to the official mirror. Same Python package, just a new hook location.

This should be the last rename here, I promise! :)


HDembinski:
> This should be the last rename here, I promise! :)

That would be great. Why don't you make PRs once things have been settled?

PR #704, opened 2022-01-04
chore: bump pre-commit

henryiii:


henryiii:
Looks like iMinuit doesn't support the latest release of MPL (3.5), though will verify (probably tomorrow or so).

PR #705, opened 2022-01-05
fix: support MPL 3.5+

### LINKED ISSUE #702: issue with 'draw_mncontour'
Hello,
In the iminuit basic_tutorial.ipynb, when I run  the given example
`m.draw_mncontour("α", "β", cl=(0.68, 0.9, 0.99))`
it leads to following error:
`ValueError: 'codes' must be a 1D list or array with the same length of 'vertices'. Your vertices have shape (101, 2) but your codes have shape (100,)`

I think it has something to do with this line
https://github.com/scikit-hep/iminuit/blob/23be056e6d25b1c27df6f0529d5d050e5ed8fb2f/src/iminuit/minuit.py#L1862

Intriguingly, the above issue happens on M1 Mac-mini. On my Macbook (intel), I don't face any such issue.

best regards

henryiii:
Closes #702 and fixes the CI on the other PRs. Matplotlib now uses closed polygons instead of unclosed lines for contours.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -1841,9 +1841,12 @@ class Minuit:
         --------
         mncontour
         """
+        import matplotlib
         from matplotlib import pyplot as plt
         from matplotlib.path import Path
         from matplotlib.contour import ContourSet
+
+        mpl_version = tuple(int(x) for x in matplotlib.__version__.split("."))
 
         cls = cl if isinstance(cl, Iterable) else [cl]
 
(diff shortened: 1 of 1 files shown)


henryiii:
This works except for coverage. @HDembinski what would you like to do for coverage?

HDembinski:
Thanks, this seems fine. I will fix the coverage in another PR.

HDembinski:
Your fix did not work, the contour had a gap. I fixed it in #699

PR #706, opened 2022-01-06
clean makefile, unpin 4.1.0 for sphinx

henryiii:
Two quick cleanups I noticed while working on recent PRs.

The Makefile can be made a bit cleaner now that python2 is no longer supported. I'd recommend looking into nox, as well; it can be much more reproducible, handing virtual environments for you. (I also hate make, so I'm biased).

I also ran into a problem with Sphnix - this pins 4.1.0 exactly, which doesn't support Python 3.10, so I couldn't build docs. You probably intended `==4.1.*`, but that's still not ideal; the 4.1 series is  not an LTS, so things like Python 3.10 support are not backported to it. I would either use `~=4.1`, which will allow 4.1 or newer, but not 5, or I'd use >=4.1, which is what I did here - [Pinning is generally a misused and terrible practice](https://iscinumpy.dev/post/bound-version-constraints). But up to you, docs is something you can pin. But I'd do at most `~=4.1`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,17 @@
 # Makefile for developers with some convenient quick ways to do common things
 
-ifeq ($(shell python -c 'import sys; print(sys.version_info.major)'), 3)
-	PYTHON = python
-else
-	PYTHON = python3
-endif
-
 # default target
 build/done: $(wildcard *.py src/*.cpp extern/root/math/minuit2/src/*.cxx extern/root/math/minuit2/inc/*.h) CMakeLists.txt
-	DEBUG=1 $(PYTHON) setup.py develop
+	DEBUG=1 python3 setup.py develop
 	touch build/done
 
 test: build/done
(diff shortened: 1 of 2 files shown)


HDembinski:
I don't hate make, because it is simple and great.

henryiii:
It is a weird, custom language that makes tabs and spaces fundamentally different, it encourages platform specific code, and it is almost never the right solution for any problem - it's _way_ overused. Need to work with Python? Nox does the environments for you - make doesn't understand environments. Need to build LaTeX? A dedicated tool like latexmk will get the reruns right. Need to build code? A dedicated tool like CMake will actually work on different OS's, will support all sorts of other tooling like IDE's, etc.

If it really, truly is the right tool for the job, then rake (real Ruby syntax) or invade (real Python syntax) are still superior.

Not saying you or anyone else can't use make, just saying why I don't like it at all.

PR #707, opened 2022-01-06
include binaries for Python 3.10, musllinux

henryiii:
I've added 3.10 & musllinux (and probably PyPy 3.8 binaries too, forgot to check to see if PyPy binaries are included). Manylinux2010 is produced on 3.6-3.9, and 2014 for 3.10, just like NumPy. @HDembinski, is there a way to trigger the wheel build without an upload to make sure it works?

Also bumped the CI to include 3.10 final.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,11 +16,11 @@ jobs:
         include:
           # version number must be string, otherwise 3.10 becomes 3.1
           - os: windows-latest
-            python-version: "3.10.0-rc.1"
+            python-version: "3.10"
           - os: macos-latest
             python-version: "3.6"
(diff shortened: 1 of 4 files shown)


HDembinski:
You can run the workflows manually, but only on the master. I will test it on the next release, which will happen ASAP to get the matplotlib fix out.

henryiii:
Okay, ping me if anything goes wrong that's not an easy fix!

HDembinski:
Correction: the manual trigger also works on develop. You could go to "Actions", choose "Wheels" and then trigger on "develop". But I am doing that right now on master in preparation for the release, so nothing to do for you. Thanks.

PR #708, opened 2022-01-06
Fix wheels

HDembinski:


HDembinski:
@henryiii It looks like there are no binary packages for numpy in musllinux, so the wheels fail:

https://github.com/scikit-hep/iminuit/runs/4732176689?check_suite_focus=true#step:4:749

HDembinski:
It makes no sense to provide binary wheels for iminuit when the corresponding binary wheels for numpy do not exist, so I will remove the wheels for musllinux. iminuit is not a test case for cibuildwheels.

henryiii:
Okay, sadly those are not available (yet). It works in boost-histogram because there NumPy is not required when building; someone on musllinux can use apk to install numpy if they want.

pyproject.toml
```diff
@@ -17,7 +17,6 @@ testpaths = [
 ]
 
 [tool.cibuildwheel]
-before-build = "python -m pip install numpy --only-binary=:all:"
 test-requires = "pytest"
 test-command = "python -m pytest {package}/tests"
 test-skip = ["*universal2:arm64"]
```
henryiii, line 22:
```suggestion
test-skip = ["*universal2:arm64"]
skip = ["*musllinux*"]
```

This can be used to skip musl.

pyproject.toml
```diff
@@ -25,5 +25,7 @@ manylinux-i686-image = "manylinux2014"
 
 [[tool.cibuildwheel.overrides]]
 select = "cp3?-*"
+# update skip when numpy wheels become available
+skip = ["*-musllinux_*", "cp310-win32"]
```
henryiii, line 29: This is an overrides section, just like mypy overrides, so it only applies on things that match the select statement, `cp3?-` (which is Python 3.6, 3.7, 3.8, and 3.9) - the skip should be placed a few lines above.

(The purpose of the override in this case is to use manylinux2010 on older Pythons, but leave 2014 for 3.10 and newer)

HDembinski: Ok.

henryiii: If you find `select = "cp3{6,7,8,9}-*"` clearer, feel free to use that instead.

henryiii: Or add comments, I didn't notice and copy over the commend in boost-histogram: https://github.com/scikit-hep/boost-histogram/blob/0ddcc8db4007978f44051a8a32ef44fa23528689/pyproject.toml#L94

.github/workflows/wheels.yml
```diff
@@ -90,6 +90,7 @@ jobs:
   upload:
     needs: [wheels, sdist]
     runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/master' }}
```
henryiii, line 93: Thanks! Something like this is what I was wanting. I could test this before making the PR as long as I'm not on master now. (You could also make sure it's coming from scikit-hep/iminuit, but now that you have this check, that's enough, at least for me)

henryiii:
FYI, the readme still says "supported PyPy versions 3.6". 3.6 has been dropped by PyPy, the "main" PyPy is 3.7 and 3.8 is also available, though still considered beta. The latest pybind11 supports 3.7 & 3.8, though we've found and worked around some of the beta bugs in 3.8. ;)

HDembinski:
Thanks, I will fix this

PR #714, opened 2022-02-24
fix: remove wheel from PEP 518 requirements

henryiii:
"wheel" is included with proper version by `get_requires_for_build_wheel` from PEP 517, and isn't required for making an sdist, and might not be required in the future at all (in which case setuptools will modify `get_requires_for_build_wheel`). It's available even in setuptools 40.8.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
     "setuptools>=42",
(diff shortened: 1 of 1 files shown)


henryiii:
I'm going to run all-repos and see what that does, will reopen if this is still needed after the all-repos run.

PR #731, opened 2022-03-21
Multivariate binned

HDembinski:
This adds support for a multivariate model and binned data in `BinnedNLL` and `ExtendedBinnedNLL`.


src/iminuit/cost.py
```diff
@@ -567,11 +572,23 @@ def _prepare_data(self, value):
 
         return n
 
+    def _pred(self, args):
+        X = np.meshgrid(*self._xe) if self._ndim > 1 else self._xe
```
maxnoe, line 576: Is recreating `X` here for every iteration expensive?  Maybe create it once in `__init__`?

src/iminuit/cost.py
```diff
@@ -567,11 +572,23 @@ def _prepare_data(self, value):
...
+        cdf = self._model(X, *args)
+        cdf = _normalize_model_output(cdf)
+        if self._ndim == 1:
+            d = np.diff(cdf)
+        elif self._ndim == 2:
+            d = cdf[1:, 1:] - cdf[1:, :-1] - cdf[:-1, 1:] + cdf[:-1, :-1]
+        else:
+            raise ValueError("general ND case with D > 2 currently not supported")
```
maxnoe, line 584: This error could already be raised in `__init__` on checking `_ndim`, right?

src/iminuit/cost.py
HDembinski: Yes, it is, good point. I also found a few other places where the code is not caching results properly that do not vary a lot. The cost functions need a redesign in regard to performance. The class hierarchy is also leaking abstractions everywhere.

src/iminuit/cost.py
HDembinski: Yes, but the `if` here is comparably cheap, so it is ok. The new version does not raise anymore.

PR #740, opened 2022-03-28
Missing test files in SDist

henryiii:
Fixes conda-forge (and any other downstream packagers running tests)

I think only html files were added?

We should add the check-manifest check, as described in the Scikit-HEP developer guidelines.


HDembinski:
Yes, that's all. Thanks for the patch.

PR #743, opened 2022-03-28
Removed unused and deprecated Black notebook config

henryiii:
Editted: I mistook the usage of tool.nbqa.mutate black=1 to mean that the notebooks were supposed to be black formatted. Instead, I'm now just removing that old setting (which is deprecated, since black now natively supports jupyter notebook formatting).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,9 +5,6 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
-[tool.nbqa.mutate]
-black = 1
-
 [tool.mypy]
 ignore_missing_imports = true
 


HDembinski:
I don't want auto-formatting for my notebooks.

pyproject.toml
```diff
@@ -5,9 +5,6 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
-[tool.nbqa.mutate]
-black = 1
```
henryiii, line 9: I was assuming you wanted auto formatting because I saw this construct, which is deprecated, so I replaced it with the alternative. (I was wondering why there were quite a few notebook changes, but most of them seems to be empty space at the end of lines)

henryiii:
Reopened this with just the removal of the confusing line in the pyproject.toml that configures `nbqa` with `black`, since a) it's deprecated and b) the notebooks are not auto formatted anyway.

pyproject.toml
HDembinski: This is a left-over of that PR that I haven't cleaned up.

PR #747, opened 2022-04-09
hide typing module in minuit.py

HDembinski:


henryiii:
FYI, if you place an `__all__ = [...]` in a module listing all your public API, then add the following function:

```python
def __dir__() -> List[str]:
    return __all__
```

That will hide everything from tab completion (even in the vanilla REPL) or `dir` except what is explicitly listed as public API via `__all__` (as long as the user has Python 3.7+). I wish I'd had/known about this when I wrote boost-histogram, I probably would not have used the `_internal` modules and it would have been simpler.

PR #752, opened 2022-05-13
Added bblite cost function

AhmedAbdelmotteleb:
@HDembinski 
Hello! So, I added my method for the BBLite function. I also included an extensive docstring if you want to review the math behind this. Feel free to edit it or remove it altogether (I saw some of the cost functions had comments explaining the algebra). Also, feel free to change the name, or anything you want really :)


HDembinski:
@AhmedAbdelmotteleb Thank you so much, this is perfect. I can write the infrastructure around this. The comment is also very helpful. Is there a reference to cite for Barlow-Beeston lite (BBL)?

I was not familiar with the lite version before, but I studied the Barlow-Beeston (BB) paper and tried to reproduce the results. BB is directly derived from first principles, while BBL seems ad hoc. Is there a way to derive BBL from BB? This would help me to understand what kind of limit we are looking at, i.e. under what conditions this is applicable.

In the BB paper, they show that the method works very well, negligible bias and variance is correctly predicted. This is as expected, since we know that ML fits with Poisson statistics work very well. Are there similar studies for Barlow-Beeston lite? The nuisance parameter is introduced as a multiplicative factor, which is constrained assuming a normal distribution, and this could introduce bias in bins with small statistic in the template.

HDembinski:
I think the BBL limit describes the approximation for large statistic in the MC template (Poisson replaced with Gaussian) and potentially low statistic in the data (data is using Poisson cost function). In practice, it often the other way round, because generating MC is so expensive.

AhmedAbdelmotteleb:
Hello @HDembinski! Apologies for the delay in response :) I have gathered multiple resources for you to review at your leisure:

- https://arxiv.org/pdf/1103.0354.pdf [reference for BBL + "derivation"]
- https://indico.cern.ch/event/107747/contributions/32677/attachments/24367/35056/Conway-PhyStat2011.pdf  [presentation  by same person from previous resource + nice plots with more explanation]
- https://w3.iihe.ac.be/~pvanlaer/RooStats/statistics_2012_part3_v9.pdf [Presentation comparing BB and BBL]
- https://www.precision.hep.phy.cam.ac.uk/wp-content/people/mitov/lectures/GraduateLectures/Advanced-Statistics-Verkerke.pdf [presentation comparing BB and BBL + commenting on accuracy of BBL]
- https://arxiv.org/pdf/1708.07708.pdf [deriving BBL from BB (slightly condensed)]

Hope these are helpful :) With regards to your questions, yes, I think BBL struggles in ultra-low MC stats: likely due to Gaussian
assumption in scaling parameters. BBL works on the assumption that for high MC event counts we can model the uncertainty with a Gaussian distribution. Do let me know if there is anything else I can do to help.

HDembinski:
@AhmedAbdelmotteleb

Thank you for the references, it is a bit too much to read, it seems the last reference is what I wanted to know.

I understand in general how one can add nuisance parameters to log-likelihood fit (this is described by the first reference for instance). However, introducing the multiplicative nuisance factor seems ad hoc and not fundamentally justified, but perhaps I am missing something. I have to read the last reference in detail to understand.

Ref https://www.precision.hep.phy.cam.ac.uk/wp-content/people/mitov/lectures/GraduateLectures/Advanced-Statistics-Verkerke.pdf
is quite long, where exactly do I find the comparison between BB and BBL?

HDembinski:
https://arxiv.org/pdf/1103.0354.pdf does not give a derivation as I would like. The multiplicative factor with a Gaussian distribution is introduced ad hoc (see discussion that leads up to Eq. 8), it is not derived from the original constraint of a Poisson distribution.

HDembinski:
These references are super useful, thank you. I am just trying to squeeze as much knowledge out of you as I can so that I have less to do ;).

I am now going to try to finish the implementation.

src/iminuit/cost.py
```diff
@@ -106,7 +106,38 @@ def _soft_l1_loss(z_sqr):
...
+    One can then solve the quadratic equation and obtain the correct (maximal/positive)root.
+        
+        ==> b = -(mu*s^2-1)/2  +/- sqrt( (mu*s^2-1)^2 - 4n*s^2 ) /2
+                = -    x /2    +/-  sqrt ( x^2       - 4n*s^2 ) /2    (3)
+                
+                ,where x = (mu*s^2-1).
+    """
+    s2 = 1/mu
+    x = mu * s2 - 1
+    b = -x * 0.5 + np.sqrt(x**2 + 4 * n * s2) * 0.5
```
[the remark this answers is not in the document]
HDembinski: Ok, I think I understand this. x is zero for poisson errors, which you assume here, but not in general. I started to implement the more general case. No need to edit your code.

src/iminuit/cost.py
```diff
@@ -107,6 +79,141 @@ def _soft_l1_cost(y, ye, ym):
...
+    lnP = sum_i -mu_i + k_i ln(mu_i) - ln(k_i!)
+    lnP' = sum_i -k_i + k_i ln(k_i) - ln(k_i!)
+    Q = 2 sum_i mu_i - k_i + k_i (ln(k_i) - ln(mu_i))
+    """
+    n = np.atleast_1d(n)
+    mu = np.atleast_1d(mu)
+    return 2 * np.sum(mu - n + n * (_safe_log(n) - _safe_log(mu)))
+
+
+def barlow_beeston_lite_chi2(n, mu, mu_var):
```
HDembinski, line 181: @AhmedAbdelmotteleb I slightly generalised and documented your function, could you please check whether this is correct?

src/iminuit/cost.py
```diff
@@ -676,6 +789,77 @@ def __init__(self, n, xe, model, verbose):
...
+            f = a / nt
+            mu += f * t
+            var_t = t  # different if weighted templates are used, TODO
+            mu_var += var_t * f**2
+
+        ma = self.mask
+        if ma is not None:
+            mu = mu[ma]
+            mu_var = mu_var[ma]
+        return barlow_beeston_lite_chi2(self._masked, mu, mu_var)
```
HDembinski, line 860: @AhmedAbdelmotteleb To apply BBL, I need to compute the mu and its variance. `mu` is the sum over the component fraction times the template value in a bin, so `mu_var`, the variance of `mu`, needs to be computed from the Poisson uncertainties of the template weighted by the component fraction. Can you please check that this calculation agrees with how you do it?

HDembinski:
@AhmedAbdelmotteleb I tried the BBL method on a toy example, but I do not get satisfactory results. I am not sure whether this is because of the approximations in BBL or whether my implementation has a bug. 

You can see the notebook here. I first compute the result using the correct likelihood like the Barlow-Beeston and then use BarlowBeestonLite. I get different best-fit values and also different uncertainties. If I send you the histograms that I generate here, could you please apply your implementation on them so that we can compare the values? This would help to find out whether my implementation has a bug.

https://nbviewer.org/github/scikit-hep/iminuit/blob/ef5bed8e5a63e5909b295df7815af596577e6647/doc/notebooks/template_fits.ipynb

HDembinski:
@matthewkenzie FYI

src/iminuit/cost.py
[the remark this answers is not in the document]
HDembinski: Having thought more about this, I think this is not correct, because `mu` is a mixture of different templates, each with a different statistical error. You need to perform error propagation to get the uncertainty of `mu` based on the uncertainties of the templates and the component fractions.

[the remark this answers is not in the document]
AhmedAbdelmotteleb: Hi @HDembinski. Apologies for the delayed response. I checked out this out and had a discussion with my supervisor. We are assuming a Poisson error. However, we were supposed to use the relative uncertainty, instead we just used the normal standard deviation (1/sqrt(N)), which is incorrect. Additionally, as you said, `mu` is a mixture of templates so one needs to determine the corresponding uncertainties and evaluate the value for `s`, where `s = uncertainties/templates`, so `s2 = (uncertainties/templates)^2 = variances/templates^2`

src/iminuit/cost.py
AhmedAbdelmotteleb: Yes, that is correct. I will check the calculation now.

AhmedAbdelmotteleb: @HDembinski Your implementation looks correct to me. I will add some suggestions though, hope that's okay :)

src/iminuit/cost.py
```diff
@@ -676,6 +787,77 @@ def __init__(self, n, xe, model, verbose):
...
+            1: print current args and negative log-likelihood value.
+        """
+        n = np.atleast_1d(n)
+        templates = np.atleast_2d(templates)
+        M = len(templates)
+        if M < 1:
+            raise ValueError("at least one template is required")
+        if n.shape != templates.shape[1:]:
+            raise ValueError("shapes of n and templates do not match")
+        self._templates = templates
```
AhmedAbdelmotteleb, line 830: I would add a scaling of the templates to the data. The way we did it is like follows:

```python
sum_n = np.sum(n)
sum_templates = np.sum(templates, axis = 1)
inv_frac = (sum_n / sum_templates) 
#might need to reshape this array to make sure multiplication is handled correctly
#inv_frac = (sum_n / sum_templates).reshape((len(templates), 1))
self.__variances = inv_frac * inv_frac * np.array(templates)
self.__templates = inv_frac * np.array(templates) 
```
and this would mean you don't need to worry about scaling later on.

src/iminuit/cost.py
```diff
@@ -676,6 +787,77 @@ def __init__(self, n, xe, model, verbose):
...
+    @Cost.ndata.getter  # type:ignore
+    def ndata(self):
+        """See Cost.ndata."""
+        shape = self._masked.shape
+        return np.prod(shape)
+
+    def _call(self, args):
+        mu = 0
+        mu_var = 0
+        for a, t, nt in zip(args, self._templates, self._sum_templates):
```
AhmedAbdelmotteleb, line 848: In continuation of my previous comment on line 830
I would use numpy arrays here instead of for loops:
```python
#might need to add a new axis to enable multiplication to work properly 
#args = np.asarray(args) 
args = np.asarray(args) [:, np.newaxis]
mu = args * self.__templates
mu_var = self.__variances * args**2
```
I think if you provide the templates normalised to the data as in the previous comment (line 830), you would not need to normalise the fit fractions/args with respect to the templates anymore.
And with numpy arrays you get faster calculations + it's easier to understand

src/iminuit/cost.py
```diff
@@ -107,6 +79,141 @@ def _soft_l1_cost(y, ye, ym):
...
+    """
+    # b_var is trivially 1 for Poisson, but not for weighted templates
+    b_var = mu_var / mu**2
+
+    # need to solve quadratic equation b^2 + (mu var_b - 1) b - n var_b = 0
+    p = mu * b_var - 1
+    q = -n * b_var
+    beta = 0.5 * (-p + np.sqrt(p**2 - 4 * q))
+
+    return poisson_chi2(n, mu * beta) + np.sum((beta - 1) ** 2 / b_var)
```
AhmedAbdelmotteleb: @HDembinski  I think the bug might come from here. 

What I return is the sum of the negative log likelihood, and I do it as follows:
```python
nll = -n * np.log(beta * mu) + beta * mu + 0.5 * (beta - 1.)**2 / b_var
return np.sum(nll)
```
, which is just equation 10 from the [main reference](https://arxiv.org/pdf/1103.0354.pdf).
I don't think it matches up with what you provide here.
As far as I can tell, you return this:
```python
return 2 * np.sum(mu*beta - n + n * (np.log(n) - np.log(mu*beta))) +  np.sum((beta - 1) ** 2 / b_var)
```
The first part of the expression (`poisson chi2`) contains log(n), which is not present the referenced formula,
and in the second part of the expression, you do not divide by 2. Please let me know what you think :)

src/iminuit/cost.py
HDembinski: I understand your point, but I intentionally do use that here. It is not significantly faster in this case and `args = np.asarray(args) [:, np.newaxis]` is not beautiful code.

src/iminuit/cost.py
HDembinski: No, that's all ok, you can add constants to the likelihood without changing anything. poisson_chi2 computes 2 * (-log(L)) and therefore I do not divide by 2 in the Gaussian constraint. It is consistent.

HDembinski:
@AhmedAbdelmotteleb 

Here is a gzipped pickle with the data files that I used in my toy study. When you load this you get a tuple. The first item is the bin edges, the second is the data histogram and the third item is a 2D numpy array that holds the templates.

[data.pkl.gz](https://github.com/scikit-hep/iminuit/files/8787856/data.pkl.gz)

src/iminuit/cost.py
AhmedAbdelmotteleb: fair enough

src/iminuit/cost.py
AhmedAbdelmotteleb: I see. Makes sense :)

AhmedAbdelmotteleb:
> @AhmedAbdelmotteleb
> 
> Here is a gzipped pickle with the data files that I used in my toy study. When you load this you get a tuple. The first item is the bin edges, the second is the data histogram and the third item is a 2D numpy array that holds the templates.
> 
> [data.pkl.gz](https://github.com/scikit-hep/iminuit/files/8787856/data.pkl.gz)

Hi @HDembinski ,
I used these files and I get these results (with a single warning that the log received an invalid value - I followed it up and found that one of the values of `mu` becomes negative, so the log complains, but then it recovers and converges properly): 
image


It seems that I get the similar results but with a factor of 1000 off your results, not sure why that is. I also printed out the correlation using the for loop you had in the notebook.

Also, I plotted the contour you had in the notebook and got this:
image

I have attached my script in case you want to try something yourself:
[test_bblite.zip](https://github.com/scikit-hep/iminuit/files/8796590/test_bblite.zip)

Let me know if you need anything else :)

src/iminuit/cost.py
HDembinski: Perfect, thanks.

src/iminuit/cost.py
HDembinski: I will probably follow this suggestion, but let's make sure first that the implementation is correct and produces reasonable results.

HDembinski:
> It seems that I get the similar results but with a factor of 1000 off your results, not sure why that is. I also printed out the correlation using the for loop you had in the notebook. 

Thank you for that. The factor of 1000 arises because I am not fitting fractions but yields, it is fraction times total number of events. Based on the way how the likelihood is defined, fitting yields is correct and not fitting fractions, actually (to obtain true fractions, one has to use a log-likelihood based on the binomial distribution, not Poisson).

So for reference, I got this in my notebook, which is indeed identical up to a factor 1000 (= ndata).

image

In other words, at least in this case toy example, the BBL method produces a result which is rather different from BB, especially the uncertainties are much larger. I will check the bias and variance of both estimators next to see which one performs better.

Issue #753, opened 2022-05-16
Tests do not work on i386 due to use of float128


jeremysanders:
On the Linux i386 platform (at least), the data type `numpy.float128` does not exist. This causes the test `test_cost.py` to fail. I believe `float128` is being used to map to `long double` in C++. https://github.com/numpy/numpy/issues/10288, suggests that the data type `numpy.longdouble` should likely be used instead of `float128`, as `long double` is not an ieee128 datatype on that platform.

See https://gist.github.com/jeremysanders/9aca53cfccd5414d43f9f33eae911287 for a patch.

HDembinski:
Thank you for this bug report, but I think the correct treatment is to skip this test on platforms that do not support this data type. Is there a way to check whether numpy supports numpy.float128 that can be used as a requirement for the test?

jeremysanders:
I don't agree, as the test works anywhere fine if the data type is changed. The C data type is long double, so why not use the correct data type in iminuit? There's no guarantee that a long double is a float182. In addition, the presence of float128 doesn't mean that a long double has this same data type.

HDembinski:
The test is explicitly for the type float128. A long double is not guaranteed to have 128 bits.

PR #754, opened 2022-05-16
Skip tests that use numpy.float128 type if it not available

jeremysanders:
The numpy.float128 is not present on some platforms (e.g. i386).
To match the C++ type "long double", numpy.longdouble should be
used instead.

Fixes bug #753


HDembinski:
@jeremysanders Thank you for this patch! I think the correct approach is to skip the test instead on platforms where the type is not supported, see my comment on the issue.

HDembinski:
Is it enough to check `hasattr(numpy, "float128")?`

PR #792, opened 2022-09-20
Add Python 3.11, drop 3.6

henryiii:
Updating to add Python 3.11. For the moment, this adds all three OS's for 3.11, can be scaled back when the PR tests pass.

FYI, trying to roll out 3.11 support in Hist and iminuit's a testing dependency.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -26,7 +26,7 @@ jobs:
         submodules: true
     - uses: actions/setup-python@v4
       with:
-        python-version: 3.8
+        python-version: "3.9"
     - run: python -m pip install --upgrade pip
     - run: python -m pip install --prefer-binary -e .[test]
     - run: coverage run -m pytest
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
         submodules: true
     - uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: "3.9"
     - run: sudo apt-get install pandoc
     - run: python -m pip install --upgrade pip wheel
     - run: python -m pip install --prefer-binary -v .[doc,test]
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -17,7 +17,7 @@ jobs:
 
     - uses: actions/setup-python@v4
       with:
-        python-version: 3.8
+        python-version: "3.9"
 
     - run: python -m pip install --upgrade pip wheel
     - run: python -m pip install pre-commit
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         arch: [auto, aarch64]
-        py: [cp36, cp37, cp38, cp39, cp310, cp311]
+        py: [cp37, cp38, cp39, cp310, cp311]
         include:
           - os: macos-latest
             py: cp38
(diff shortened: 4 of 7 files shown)


HDembinski:
Why do you use iminuit in testing hist? iminuit is histogram-library agnostic.

HDembinski:
I am not interested in updating github actions when the old versions still work. If it is not broken, don't "fix" it.

henryiii:
> Why do you use iminuit in testing hist? iminuit is histogram-library agnostic.

Because `hist[plot]` has a dependency on iminuit. Some of the plot functions require iminuit.

> I am not interested in updating github actions when the old versions still work. If it is not broken, don't "fix" it.

Old versions of actions are not supported and might not support new versions of Python. v2 does not support PyPy 3.8 & 3.9, for example. I would not recommend enabling Python 3.11 using a version of actions no longer supported that was released before 3.11 was released.

v2 added the ability to target versions of Python that were not installed. v3 added the ability to target PyPy 3.8 & 3.9 (3.7 isn't supported by numpy anymore and has been discontinued in PyPy's master branch, FYI). v4 started requiring the Python version be specified. (We contributed some features to 4.1, FYI, that cibuildwheel and nox use). There were other changes too, like support for caching pip dependencies, but those were the "breaking" changes that caused the version number to bump.

In general, GitHub Actions is a moving platform. You can't control this. September 7th, all actions will start running in Node 16 instead of Node 12. You can't control software updates either - the virtual environment is updated at a set schedule (I think weekly). Actions are just a JavaScript interface to the updating backend and old versions of actions may eventually break, some features will be missing, and will be slower. GitHub does not provide v1, etc. so that you can avoid ever upgrading, they provide it so that you can chose to upgrade and not be surprised by something breaking.

And v2 was the "latest" version when you write this action, so you are only complaining because I want to update it; if you wanted to be on "older" actions, you'd have picked v1. And old version _could_ be updated. If you really want to keep it from changing, you can select an exact tag. (and then I'd _highly_ recommend using Dependabot or that will surely break when the virtual environments or Node gets updated).

Also, your current actions are broken due to matplotlib warnings.

HDembinski:
> In general, GitHub Actions is a moving platform. You can't control this.

I am well aware, but I chose to handle this differently.

henryiii:
Then please add Python 3.11 wheels.

henryiii:
Remember `python_requires = >=3.7`, too!

henryiii:
I can rebase it for you, or you can - let me know.

HDembinski:
> I can rebase it for you, or you can - let me know.

Thanks, no I will take it from here.

henryiii:
Okay, great. Don't forget to update `python_requires = >=3.6` or you will break 3.6 users instead of leaving them on a working old version.

HDembinski:
> Okay, great. Don't forget to update `python_requires = >=3.6` or you will break 3.6 users instead of leaving them on a working old version.

Good point, thanks.

henryiii:
Why did you revert pybind11 to 2.9.2? I intentionally increased it to 2.10.0 since that's the first version will Python 3.11 support.

henryiii:
(though if it's being tested, it might be okay - most of the 3.11 fixes I can recall were related to certain inheritance situations and embedding Python)

HDembinski:
> Why did you revert pybind11 to 2.9.2? I intentionally increased it to 2.10.0 since that's the first version will Python 3.11 support.

That was accidental.

henryiii:
Ah, okay. It's passing tests so I think it's fine. Fixes were related to edge cases probably not touched by IMinuit.

If you'd like to make it a bit easier in the future, you can set `git config submodule.recurse true` (or `--global` to put it into `~/.gitconfig` instead of `./.git/config` for all repos). Then git will update the submodule when you pull the parent repo and the submodule checkout commit has changed. I found out about this pretty recently and don't have a lot of first-hand experience with it, but it sounds useful. (requires Git 2.15+ or 2.16+, forget which).

PR #793, opened 2022-09-20
Add dependabot to keep actions up to date

henryiii:
Many of the actions are quite out of date; using dependabot (and I've set a monthly setting to keep the number of updates down) would really help keep the actions fresh & active. See most of the other scikit-hep repos for examples of working dependabot files. Dependabot recently became smart enough to only do "matching" updates; v2 -> v3 and not v3.0.1 or something like that.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


henryiii:
FYI, matplotlib 3.6 has rather buggy deprecation warnings - it seems to trigger them using it's own calls. Not sure that this is what is causing the failure, please check, but don't be surprised if it's unavoidable at the moment. You can just add the following line in `[tool.pytest.ini_options]:

```toml
"ignore::matplotlib._api.deprecation.MatplotlibDeprecationWarning",
```

To make it pass the tests.

HDembinski:
I still don't want a bot to update my actions which have carefully handwritten and I don't want to test always against the most recent versions of software.

henryiii:
The bot makes PRs (monthly) for major updates (it exactly matches whatever you wrote by hand). You can see if it fails or passes. You can chose to not update if you want. But actions can and will break if you go on unsupported versions of actions for too long. For example, on the 27th Node gets updated from 12 to 16, and it's quite possible the old action's won't work then. If not then, maybe the next Node update.

This does not affect the software you are running on, just the javascript actions you are using that talk to an online API.

henryiii:
FYI, GitHub is asking authors to make sure they are on the latest actions before the Node upgrade: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

eduardo-rodrigues:
Hi gentlemen. FWIW I have been using this monthly update service and it's been rather helpful, I must say. In the end it is hardly any extra work for the maintainer, and it helps having services updated. Recently, for example, such an update started provoking warnings for the move to Pandas 1.5 and it only took me a few minutes to check it out and update my code, which is now ready for 1.5. It was beneficial IMO that I got to know about the future change (all keyword arguments to a function) upfront.

As Henry says, you can always ignore the update (just close it) if you can't be bothered to act on something that requires more action than a single approve and merge. I have done that on occasion but in most cases it was an obvious accept.

My 2 cents. Your choice, obviously.

HDembinski:
Fine, you both convinced me to give it a try at least.

henryiii:
FYI, the first time this makes PRs, it will open one per dependency that needs an update (see https://github.com/scikit-build/scikit-build/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies for an example). Each PR describes exactly what has changed, with an embedded copy of the changelog and commit list. Once you are up to date, then actions releases come out slowly (~1 per year or so), so you won't have as many in-flight.

Just wanted to make sure you don't get surprised by a bunch of PRs and then think that's going to continue.

PR #799, opened 2022-09-30
added visualize function to Minuit

jsaarela1:
- Adds Minuit.visualize(), which calls visualize on the cost function with current best fit values
- Fixes Minuit.interactive() for ipywidgets >= 8.0 in Jupyter notebook (seems to be broken in VSCode now)


HDembinski:
Thanks for the patch, but please find a solution that works without importing the cost submodule.

PR #805, opened 2022-10-26
ci: update to Python 3.11 final release

henryiii:


henryiii:
The docs build seems to be hanging. That should not related to the update in this PR at all, docs run on 3.9. `¯\_(ツ)_/¯`

PR #815, opened 2022-12-14
add configurable number of bins for extended/unbinnedNLL

### LINKED ISSUE #772: UnbinnedNLL and ExtendedUnbinnedNLL: make number of bins configurable
`UnbinnedNLL` and `ExtendedUnbinnedNLL` use the `visualize` method in `UnbinnedCost`. The number of bins used in the visualization is currently fixed to 50, this should become user-configurable.

amanmdesai:
Closes #772

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -767,7 +767,7 @@ class UnbinnedCost(MaskedCost):
         # unbinned likelihoods have infinite degrees of freedom
         return np.inf
 
-    def visualize(self, args: _ArrayLike, model_points: int = 0, nbins = 50):
+    def visualize(self, args: _ArrayLike, model_points: int = 0, nbins: int = 50):
         """
         Visualize data and model agreement (requires matplotlib).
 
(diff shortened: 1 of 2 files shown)


HDembinski:
Thanks for this. It looks (it is a really simple patch :)). Some things are still missing:
- type annotation for the new parameter
- a unit test (it is a bit difficult for visualize)
The test errors are unrelated to the patch, something weird is going on in the latest Jupyter version. I don't know yet how to fix this.

HDembinski:
Thanks! The test is ok-ish, it checks that calling that interface works. It does not check whether the change has an effect on the image, but that's also true for the other tests that we are currently doing. I currently do not have the infrastructure in place to compare images produced with a reference.

Last thing to do: Could you please install pre-commit and run `pre-commit run -a` locally, to fix the formatting issues in your patch? Then it is ready to be merged.

HDembinski:
> (Unfortunately pre-commit commands dont work on my ubuntu 22 system.)

That is strange, because they pull in the dependencies in docker images that should work on any system.

HDembinski:
Thanks for the patch!

PR #839, opened 2023-02-15
Support annotated model parameters

HDembinski:
This enables support in iminuit and the builtin cost function for parameter annotations that declare parameter limits. For parameter annotations to work in Python-3.8, you should use `from __future__ import annotations`.

Parameter limits are declared with the annotated type `Annotated[float, ValueRange(min, max)]` where `min` and `max` are floats including positive or negative infinity. Annotations of this kind are recognized and automatically used to set parameter limits in `Minuit`. To silence linter warnings in Python-3.9+ you can do 
```py
from typing import Annotated
from iminuit.typing import ValueRange
```


jpivarski (approved):
This is a very nice feature! It's a good use of using types to express meaning.

HDembinski:
> This is a very nice feature! It's a good use of using types to express meaning.

Thanks, I added some more comments on my mini demo to explain why this is useful.

agoose77:
Cool suggestion Hans. Moving this metadata to the function seems like a good suggestion.

> When `from __future__ import annotations` is used, `ValueRange` does not have to be imported from iminuit.typing, the type does not have to exist at all, or could be a different type defined in another library. It is only required that ValueRange behaves like a tuple of two floats.

PEP 649 is seeming like it will be accepted at some point (the hold-up appears to be in the exact implementation details). I think, therefore, we'd need to avoid encouraging people not to import `ValueRange`, as I'm fairly sure that this would no longer work with PEP 649.

HDembinski:
Thanks @agoose77, this is why requested review from a large number of people for this change, because these annotations are still pretty new and somewhat in flux. Can you explain the implications of https://peps.python.org/pep-0649 for this feature a bit better?

The double negation here: "we'd need to avoid encouraging people not to import ValueRange" is confusing me. Can you elaborate?

eduardo-rodrigues:
This is very nice and useful! Thanks a lot for sharing 👍!

eduardo-rodrigues (approved):

henryiii:
My only comment would have been the same as @agoose77 - annotations will probably not always be strings (the exact future of `from __future__ import annotations` is unclear, but it possibly will become a no-op post PEP 649). I would avoid advertising anything special when `from __future__ import annotations` is used, and always show the `iminuit.typing` import. It's also best to not have undefined annotations anyway, as static tools will complain, runtime typing tools will break, etc. Explicitly using a string for an annotation will remain valid, so it's fine to support it, just not advertise the "future" shortcut.

src/iminuit/typing.py
```diff
@@ -39,3 +34,17 @@ class LossFunction(Protocol):
     def __call__(self, z: NDArray) -> NDArray:
         """Evaluate loss function on values."""
         ...  # pragma: no cover
+
+
+class ValueRange(NamedTuple):
```
henryiii, line 39: Personally, I'd make these Protocols, not NamedTuples. NamedTuple is a backward compatibility feature with a lot of bloat (tuple methods, for example). But don't think it really matters for this.

henryiii: Oops, forgot they need to be instantiated. 🤦 Then would go with a dataclass over NamedTuple. Again, not at all a problem.

src/iminuit/util.py
```diff
@@ -1475,3 +1561,21 @@ def _smart_sampling(f, xmin, xmax, start=5, tol=5e-3):
...
     return np.transpose(xy)
+
+
+def _get_annotations(obj: Any) -> Dict[str, Any]:
+    # Compatibility wrapper for CPython < 3.10,
+    # see https://docs.python.org/3/howto/annotations.html
+    #
+    # Returns a dict that maps parameter names to annotations.
+    # If a type is not annotated, it is ommitted.
+    try:
```
henryiii, line 1572: This should be based on `sys.version_info`, not try/except with type ignores.

henryiii (approved):

pyproject.toml
```diff
@@ -50,6 +50,7 @@ test = [
     "joblib",
     "jacobi",
     "matplotlib",
+    "numpy<1.24",
```
matthewfeickert, line 53: A question out of curiosity only, why is the upper bound on numpy needed here? I haven't been paying attention to NumPy releases lately, so are there some changes that we should be paying attention to?

matthewfeickert (approved):
Sorry for the slow reply, but LGTM! I don't have anything to add beyond what Henry and Angus already have, but I like this feature idea quite a bit!

agoose77:
> Thanks @agoose77 for the valuable input, this is why requested review from a large number of people for this change, because these annotations are still pretty new and somewhat in flux and I don't claim to understand all of it. Can you explain the implications of [peps.python.org/pep-0649](https://peps.python.org/pep-0649) for this feature a bit better?

I think @henryiii summarised my thoughts well here. On a specific note, it's not clear to me exactly how PEP 649 will be implemented (and I think from the Discourse discussion, it's not clear to the authors either), but one interpretation is that the `__annotations__` object is a descriptor that evaluates the annotations upon inspection. As such, any symbols that are never defined in a module will throw errors.

> The double negation here: "we'd need to avoid encouraging people not to import ValueRange" is confusing me. Can you elaborate?

In this particular instance I meant the double negative! "Don't tell people that they can avoid importing `ValueRange`" might be a bit clearer, though!

pyproject.toml
HDembinski: It is only required for testing and temporary, because I use numba in the tests and current numba complains when you have numpy >= 1.24.

src/iminuit/util.py
HDembinski: Why?

HDembinski:
> annotations will probably not always be strings

I will attempt to add a test that checks that the current implementation implement works both with strings and real types, to be future safe. Importing ValueRange from iminuit is not very useful though, for the intended use, because a library that wants to add annotations that work with iminuit should not have to require iminuit as a dependency.

Ideally, annotations such as ValueRange should be standardized and put in the standard lib. I searched long and far for an external library with standard annotations, but couldn't find one. The name ValueRange can be found [in the official typing docs](https://docs.python.org/3/library/typing.html), but one cannot import this from `typing`.

HDembinski:
Annother caveat of these annotations is that they don't propagate automatically.

```
NonNegative = Annotated[float, ValueRange(0, np.inf)]

# signal and background may be reusable functions from a third-party library like numba stats
def signal(x, n: NonNegative, mu, sigma: NonNegative): ...
def background(x, a: NonNegative, b: NonNegative): ...

# if the user wants to have a model which mixes signal and background, they have to annotate again
def mixture(x, n: NonNegative, mu, sigma: NonNegative, a: NonNegative, b: NonNegative):
     return signal(x, n, mu, sigma) + background(a, b)
```

This can be solved with a reusable function `make_mixture_model` that performs this propagation of annotations automatically, but somehow I wish there was a more elegant way.

HDembinski:
> In this particular instance I meant the double negative! "Don't tell people that they can avoid importing ValueRange" might be a bit clearer, though!

Ok, thanks. In the meantime, I read [pep-0649](https://peps.python.org/pep-0649) more carefully and I hope it gets through like this. The arguments brough forth against [pep-0563](https://peps.python.org/pep-0563) ring very true, it feels like a hasty awkward hack in comparison.

src/iminuit/util.py
henryiii: One of the reasons, but yes. Code is better than comments - notice you put a comment in saying that this is due to Python version. Using `sys.version_info` makes the comment unnecessary, as it's now expressed in code, and static tools (like mypy, as you noted) can use the information. Tools like pyupgrade can detect and remove dead code blocks when you upgrade, to keep this from becoming technical debt. I generally grep for `version_info` when dropping a Python version, as well, just in case pyupgrade misses some.

HDembinski: > Code is better than comments 

True.

src/iminuit/typing.py
HDembinski: Ok, I will use dataclass, I was not aware that NamedTuple from typing carries that much bloat and made this choice not thinking much about it. A dataclass is more idiomatic modern Python, I agree.

HDembinski: I will keep this convo unresolved as a reminder to self.

HDembinski: Actually, NamedTuple was correct, because I wanted tuple semantics and a dataclass does not give you that.

HDembinski:
To those which are still following this. I made some important changes after discovering the external package annotated-types.

I want the annotations to come from a third-party library, because a library like scipy.stats that provides models that can be used in iminuit's cost functions should not have to import iminuit to get annotations that work with iminuit. Experts apparently thought about this already, and the maintainers of Pydantic and Hypothesis made the third-party package [annotated-types](https://pypi.org/project/annotated-types).

iminuit supports the annotations Gt, Ge, Lt, Le from annotated-types. In addition, reading the code in annotated-types made me aware of the fact that one can declare value ranges also with slice notation, since the slice appears inside `[]` brackets, e.g. `Annotated[float, 0:]` for a positive parameter. iminuit now also supports that. Until proper standard annotations exist, I would recommend the slice notation, because it does not depend on a third-party library.

PR #858, opened 2023-04-05
add basic latex display support if unicodeit is installed

HDembinski:
If the external package [unicodeit](https://github.com/svenkreiss/unicodeit) is installed, translate variable names which contain simple LaTeX into unicode that mimics the correct LaTeX rendering. This only works for very basic LaTeX like `$\alpha$`.


HDembinski:
unicodeit could in principle render more complex LaTeX code, if it used a better LaTeX parser. The current code just does simple search and replace, and cannot handle code such as `$D^{\ast\ast}$` correctly, since it does not understand the grouping with curly braces.

zjkjsd:
Indeed! This is awesome! Thank you for adding this feature~

Issue #791, opened 2022-09-12
Improve error messages when using builtin cost functions


HDembinski:
There are many ways to make mistakes and use the builtin cost functions incorrectly. The user has to provide the model and it is easy to return an array of incorrect size.

```py
from iminuit import cost, Minuit
from numba_stats import norm
import numpy as np

n = [1, 2, 3]
edges = [0.1, 0.2, 0.3, 0.4]

def model(x, n, a, b):
    # oopsi, returned probability per bin instead of scaled cdf
    return n * np.diff(norm.cdf(x, a, b))
    # return n * norm.cdf(x, a, b)  # correct

c = cost.ExtendedBinnedNLL(n, edges, model)
c(1, 2, 3)
```
This produces a very confusing error message:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/var/folders/tl/pv6mt7z17tz0stm1fjfg01cc0000gn/T/ipykernel_28667/1963231075.py in <module>
     12 
     13 c = cost.ExtendedBinnedNLL(n, edges, model)
---> 14 c(1, 2, 3)

/usr/local/lib/python3.8/site-packages/iminuit/cost.py in __call__(self, *args)
    482         float
    483         """
--> 484         r = self._call(args)
    485         if self.verbose >= 1:
    486             print(args, "->", r)

/usr/local/lib/python3.8/site-packages/iminuit/cost.py in _call(self, args)
   1292         else:
   1293             n = self._masked
-> 1294         return poisson_chi2(n, mu)
   1295 
   1296 

/usr/local/lib/python3.8/site-packages/iminuit/cost.py in poisson_chi2(n, mu)
    345         n, mu = np.atleast_1d(n, mu)  # type:ignore
    346         if mu.dtype in (np.float32, np.float64):  # type:ignore
--> 347             return _poisson_chi2_nb(n, mu)
    348         # fallback to numpy for float128
    349         return _poisson_chi2_np(n, mu)

ValueError: unable to broadcast argument 1 to output array
File "/usr/local/lib/python3.8/site-packages/iminuit/cost.py", line 1, 
```
We need a better error message for this cost function and for others.

amanmdesai:
It seems this issue did not close automatically after the merge of #863 (as i did not included close/fix in the message).

PR #863, opened 2023-04-10
improve error message for cost function

amanmdesai:
Hello @HDembinski

This PR is regarding https://github.com/scikit-hep/iminuit/issues/791

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -1040,7 +1040,6 @@ class BinnedCost(MaskedCost):
 
         n = _norm(n)
         is_weighted = n.ndim > self._ndim
-
         if n.ndim != (self._ndim + int(is_weighted)):
             raise ValueError("n must either have same dimension as xe or one extra")
 
@@ -1141,8 +1140,6 @@ class BinnedCostWithModel(BinnedCost):
             self._model_xe = np.row_stack(
                 [x.flatten() for x in np.meshgrid(*self.xe, indexing="ij")]
             )
-            if len(self._xe_shape) != len(self._model_xe):
-                raise ValueError("Variable shapes do not match")
 
     def _pred(self, args: Sequence[float]) -> NDArray:
         d = self._model(self._model_xe, *args)
@@ -1154,10 +1151,10 @@ class BinnedCostWithModel(BinnedCost):
         # differences can come out negative due to round-off error in subtraction,
         # we set negative values to zero
         d[d < 0] = 0
-        if len(self._data) != len(d):
+        if self._data.shape != d.shape:
             raise ValueError(
-                f"Expected model to return an array of size {len(self._data)},"
-                f"but it returns an array of size {len(d)}"
+                f"Expected model to return an array of shape {self._data.shape}, "
+                f"but it returns an array of shape {d.shape}"
             )
         return d
 
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -1126,12 +1126,12 @@ class BinnedCostWithModel(BinnedCost):
     :meta private:
     """
 
-    __slots__ = "_xe_shape", "_model", "_model_xe"
+    __slots__ = "_xe_shape", "_model", "_model_xe", "_n"
 
     def __init__(self, n, xe, model, verbose):
         """For internal use."""
         self._model = model
-
+        self._n = n
         super().__init__(_model_parameters(model), n, xe, verbose)
 
         if self._ndim == 1:
@@ -1153,6 +1153,11 @@ class BinnedCostWithModel(BinnedCost):
         # differences can come out negative due to round-off error in subtraction,
         # we set negative values to zero
         d[d < 0] = 0
+        if len(self._n) != len(d):
+            raise ValueError(
+                f"Expected model to return an array of size {len(self._n)},"
+                f"but it returns an array of size {len(d)}"
+            )
         return d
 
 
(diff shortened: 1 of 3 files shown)


HDembinski:
Cool, thanks! I think the error message is spot on.

Would you be willing to write a test for this? Otherwise I would merge as is and write a test at some later point.

A test would go to `tests/test_cost.py` and the pytest facility to use is `with pytest.raises(ValueError): <create condition which triggers the exception>`. Please have a look at examples in the source code, just search for `pytest.raises`.

HDembinski:
Thinking more about it, there are a couple of things that need to be improved.

- I think we should check that the shapes are compatible, not only the length, since the binned cost functions support N-dimensional histograms.
- The check should be implemented in class `BinnedCostWithModel`, so that all derived classes benefit from it. A good place seems to be `BinnedCostWithModel._pred`.

amanmdesai:
Hi @HDembinski 

I have added a test for raisevalue error using pytest.raises. However this test fails since scipy is not installed. 
In some cases, the other tests in test_cost use the following statement:

```
@pytest.mark.skipif(not scipy_available, reason="scipy.stats is needed")
```

so shall I include the above statement to skip test if scipy is not available?
locally, the test seem to pass.

I have also implemented a check for shape compatibility. 

Please let me know if something needs to be changed.

Thanks,
Aman

HDembinski:
> so shall I include the above statement to skip test if scipy is not available?

Yes, please. This is happening because scipy is an optional dependency in some methods. Those methods work with and without scipy being installed, but execute different code in both cases. We want to test both code paths, that's why the tests are run with and without scipy. When you test a method for which scipy is required, you need to explicitly mark so that it is skipped in those test runs.

amanmdesai:
> > so shall I include the above statement to skip test if scipy is not available?
> 
> Yes, please. This is happening because scipy is an optional dependency in some methods. Those methods work with and without scipy being installed, but execute different code in both cases. We want to test both code paths, that's why the tests are run with and without scipy. When you test a method for which scipy is required, you need to explicitly mark so that it is skipped in those test runs.

I see. Thanks. I have added that statment.
I am not sure what to write in tests to check compatibility of shapes (which is probably the reason for decrease in coverage).

src/iminuit/cost.py
```diff
@@ -1126,12 +1126,12 @@ class BinnedCostWithModel(BinnedCost):
...
     """
 
-    __slots__ = "_xe_shape", "_model", "_model_xe"
+    __slots__ = "_xe_shape", "_model", "_model_xe", "_n"
 
     def __init__(self, n, xe, model, verbose):
         """For internal use."""
         self._model = model
-
+        self._n = n
```
HDembinski, line 1134: It is not necessary to add `_n` to the slots. If you follow the inheritance chain upwards to `MaskedClost`, you can see that the data array `n` is stored in the slot `_data`. So you can query `self._data` instead of `self._n`

amanmdesai: Thanks. Done.

src/iminuit/cost.py
```diff
@@ -1142,6 +1142,8 @@ def __init__(self, n, xe, model, verbose):
             self._model_xe = np.row_stack(
                 [x.flatten() for x in np.meshgrid(*self.xe, indexing="ij")]
             )
+            if len(self._xe_shape) != len(self._model_xe):
+                raise ValueError("Variable shapes do not match")
```
HDembinski, line 1145: I think this always matches by construction, no?

src/iminuit/cost.py
```diff
@@ -1153,6 +1155,11 @@ def _pred(self, args: Sequence[float]) -> NDArray:
         # differences can come out negative due to round-off error in subtraction,
         # we set negative values to zero
         d[d < 0] = 0
+        if len(self._n) != len(d):
```
HDembinski: Here you can use `self._data`.

src/iminuit/cost.py
```diff
@@ -1153,6 +1154,11 @@ def _pred(self, args: Sequence[float]) -> NDArray:
         # differences can come out negative due to round-off error in subtraction,
         # we set negative values to zero
         d[d < 0] = 0
+        if len(self._data) != len(d):
+            raise ValueError(
+                f"Expected model to return an array of size {len(self._data)},"
+                f"but it returns an array of size {len(d)}"
+            )
```
HDembinski, line 1161: Good, but `d` and `self._data` could be multidimensional, so we want to check `self._data.shape == d.shape`.

HDembinski: The error message needs to be adapted according, "array of size" -> "array with shape" etc

tests/test_cost.py
```diff
@@ -1398,3 +1398,24 @@ def test_deprecated_Template_method():
     with pytest.warns(np.VisibleDeprecationWarning):
         t = Template([1], [2, 3], [[1], [2]], method="hpd")
         t._impl is cost.template_chi2_da
+
+
+@pytest.mark.skipif(not scipy_available, reason="scipy.stats is needed")
+def test_error_message_cost():
+    from iminuit import cost
+    import numpy as np
+    from scipy.stats import norm
```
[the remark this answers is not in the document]
HDembinski: Please also test the multi-dimensional case, where `edges` is a tuple of two arrays, with edges for the x and y-axis, and n is a 2d array.

amanmdesai:
Hi @HDembinski 

I have implemented all suggesstions with the exception of 2D test (I could not implement this test).

HDembinski:
> I have implemented all suggesstions with the exception of 2D test (I could not implement this test).

Ok, I implemented the 2D test and found some issue with our implementation. We should already test the length of the array returned by the model before it is further reshaped by `_pred`, otherwise there still will be confusing error messages.

I made a couple of unrelated changes, to make mypy happier and to increase coverage back to 100 %.

amanmdesai:
> > I have implemented all suggesstions with the exception of 2D test (I could not implement this test).
> 
> Ok, I implemented the 2D test and found some issue with our implementation. We should already test the length of the array returned by the model before it is further reshaped by `_pred`, otherwise there still will be confusing error messages.
> 
> I made a couple of unrelated changes, to make mypy happier and to increase coverage back to 100 %.

Thanks a lot @HDembinski for your help and guidance!

HDembinski:
Thanks @amanmdesai !

PR #895, opened 2023-05-12
fix: make nograd not use grad at all in automatic diff doc

jonas-eschle:


HDembinski:
Thank you very much for this fix.

PR #904, opened 2023-06-20
fix: SDist size reduction

henryiii:
This reduces the size of the SDist from 160 MB to 2.2 MB (mostly docs, notebooks, and tests, I expect). (Previously this was done at https://github.com/scikit-hep/iminuit/blob/v2.21.3/MANIFEST.in)


HDembinski:
Thanks, good catch.

PR #907, opened 2023-06-23
fix test_UnbinnedNLL_visualize to use pytest.warns()

mgorny:
Fix test_UnbinnedNLL_visualize to use pytest.warns() to check for the warning instead of pytest.raises().  This fixes the test failure when the test suite is run with -Wdefault, as is the case on Gentoo.


HDembinski:
Good catch, thank you for the commit. This was just fixed by another PR from another user.

Issue #902, opened 2023-06-20
`test_Template_with_model_2D ` fails on aarch64


ggardet:
`test_Template_with_model_2D ` fails on openSUSE Tumbleweed aarch64 with:
```
[  162s] =================================== FAILURES ===================================
[  162s] _________________________ test_Template_with_model_2D __________________________
[  162s] 
[  162s]     @pytest.mark.skipif(not scipy_available, reason="scipy.stats is needed")
[  162s]     def test_Template_with_model_2D():
[  162s]         truth1 = (1.0, 0.1, 0.2, 0.3, 0.4, 0.5)
[  162s]         x1, y1 = mvnorm(*truth1[1:]).rvs(size=int(truth1[0] * 1000), random_state=1).T
[  162s]         truth2 = (1.0, 0.2, 0.1, 0.4, 0.3, 0.0)
[  162s]         x2, y2 = mvnorm(*truth2[1:]).rvs(size=int(truth2[0] * 1000), random_state=1).T
[  162s]     
[  162s]         x = np.append(x1, x2)
[  162s]         y = np.append(y1, y2)
[  162s]         w, xe, ye = np.histogram2d(x, y, bins=(3, 5))
[  162s]     
[  162s]         def model(xy, n, mux, muy, sx, sy, rho):
[  162s]             return n * 1000 * mvnorm(mux, muy, sx, sy, rho).cdf(np.transpose(xy))
[  162s]     
[  162s]         x3, y3 = mvnorm(*truth2[1:]).rvs(size=int(truth2[0] * 10000), random_state=2).T
[  162s]         template = np.histogram2d(x3, y3, bins=(xe, ye))[0]
[  162s]     
[  162s]         cost = Template(w, (xe, ye), (model, template))
[  162s]         assert cost.ndata == np.prod(w.shape)
[  162s]         m = Minuit(cost, *truth1, 1)
[  162s]         m.limits["x0_n", "x0_sx", "x0_sy"] = (0, None)
[  162s]         m.limits["x0_rho"] = (-1, 1)
[  162s]         m.migrad()
[  162s]         assert m.valid
[  162s] >       assert_allclose(m.values, truth1 + (1e3,), rtol=0.1)
[  162s] 
[  162s] tests/test_cost.py:1362: 
[...]

PR #915, opened 2023-07-26
fix safe_log on systems which use 32 bit floats

### LINKED ISSUE #913: `_safe_log` is not safe
I found that `_safe_log` is unfortunately not a safe operation in all cases. Only affected in architectures with bad floating point precision.

The follow logs are available:

- [i386 (aka x86 32-bit)](https://buildd.debian.org/status/fetch.php?pkg=iminuit&arch=arm64&ver=2.22.0-1&stamp=1688430799&raw=0): fails during `_ubinned_dll` (probably trying to sum `-inf` from `x=0`)
- [s390x](https://buildd.debian.org/status/fetch.php?pkg=iminuit&arch=s390x&ver=2.22.0-1&stamp=1688430563&raw=0): fails during `_unbinned_dll` in `test_UnbinnedNLL` (not quite clear to me where exactly this goes wrong)
- [mips64el](https://buildd.debian.org/status/fetch.php?pkg=iminuit&arch=mips64el&ver=2.22.0-1&stamp=1688432984&raw=0): fails during `poisson_chi2`
- [mipsel](https://buildd.debian.org/status/fetch.php?pkg=iminuit&arch=mipsel&ver=2.22.0-1&stamp=1688439302&raw=0): fails during `poisson_chi2`

What connects all of these failures is the fact that they call `_safe_log`. Notably, the implementation:
```python3
def _safe_log(x):
    # guard against x = 0
    return np.log(x + 1e-323)
```

`1e-323` was probably picked since it is close to `np.finfo(float).smallest_subnormal` (which would be the exact value).

However, this only works when working with 64-bit floats, `np.finfo(np.float32).smallest_subnormal` evaluates to `1e-45`. I think the issue is that during the addition when x is a 32-bit float, `1e-323` is casted to a 32-bit float for some reason. Because ` np.float32(1e-323)` evaluates to `0.0` the log returns `-inf`.

One way this could be tested more precisely is by checking the result of `_safe_log` in the tests is actually not `-inf` since `np.log(0.0)` does not throw any error (would need to be adjusted in `test_without_numba`).

I'm not entirely sure what the best solution is here - I see 3 options:
1. Instead of adding something to x, don't execute the log for values where x is zero: `np.log(x, where=(x>0.), out=np.full_like(x, np.log(np.finfo(float).smallest_subnormal)))`. Disadvantage: always returns array even for basic floats, potentially slow
2. Use the 32-bit smallest subnormal instead of the 64-bit one. Disadvantage: potential loss of precision.
3. Provide a `_safe_log` option that depends on the datatype and uses the appropriate smallest subnormal. However I'm not sure how this can be implemented.
4. Clip the `-inf`: `np.clip(np.log(x), -1e3, None)`. Disadvantage: slightly slower than current implementation probably.

I did a quick test of option 2 using the following code:
```python3
import numpy as np

_SSB32 = np.finfo(np.float32).smallest_subnormal
_SSB64 = np.finfo(np.float64).smallest_subnormal
_reasonably_small = 1e-40
_normal_small = 1e-30

def pretty(x):
  return np.format_float_scientific(x, exp_digits=3, precision=4, min_digits=4)

def logdiff(x):
  return np.log(x+_SSB32)-np.log(x)

print(f"""
For {pretty(_SSB64)}: {pretty(logdiff(_SSB64))}
For {pretty(_SSB32)}: {pretty(logdiff(_SSB32))}
For {pretty(_reasonably_small)}: {pretty(logdiff(_reasonably_small))}
For {pretty(_normal_small)}: {pretty(logdiff(_normal_small))}
""")
```
Output:
```
For 4.9407e-324: 6.4116e+002
For 1.4013e-045: 6.9315e-001
For 1.0000e-040: 1.4013e-005
For 1.0000e-030: 0.0000e+000
```

So I don't think it's entirely unreasonable, for more or less all number people should use (i.e. x>1e-30), the difference is precisely zero. However, for the current approach this limits is roughly 1e-310.

I think overall option 4 provides the best balance between precision, performance, ease of implementation and platform compatibility.

HDembinski:
Closes #913 see that issue for a detailed analysis by @stephanlachnit why the original naive code did not work everywhere.

Maybe this is related to #902, too.

@stephanlachnit Could you please check whether this fix indeed solves the issue? I cannot run tests on the affected systems.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -116,7 +116,7 @@ NEGATIVE_LOG_LIKELIHOOD = 0.5
 
 def _safe_log(x):
     # guard against x = 0
-    return np.log(x + 1e-323)
+    return np.log(np.maximum(np.finfo(x.dtype).smallest_normal, x))
 
 
 def _unbinned_nll(x):
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -113,10 +113,12 @@ __all__ = [
 CHISQUARE = 1.0
 NEGATIVE_LOG_LIKELIHOOD = 0.5
 
+_SMALLEST_NORMAL = np.finfo(float).smallest_normal
+
 
 def _safe_log(x):
     # guard against x = 0
-    return np.log(np.maximum(np.finfo(x.dtype).smallest_normal, x))
+    return np.log(np.maximum(_SMALLEST_NORMAL, x))
 
 
 def _unbinned_nll(x):
--- a/tests/test_cost.py
+++ b/tests/test_cost.py
@@ -1125,7 +1125,7 @@ def test_multinominal_chi2():
 
     assert multinominal_chi2(zero, zero) == 0
     assert multinominal_chi2(zero, one) == 0
-    assert multinominal_chi2(one, zero) == pytest.approx(1487, abs=1)
+    assert multinominal_chi2(one, zero) == pytest.approx(1416, abs=1)
     n = np.array([(0.0, 0.0)])
     assert_allclose(multinominal_chi2(n, zero), 0)
     assert_allclose(multinominal_chi2(n, one), 0)


src/iminuit/cost.py
```diff
@@ -116,7 +116,7 @@ class documentation for details.
 
 def _safe_log(x):
     # guard against x = 0
-    return np.log(x + 1e-323)
+    return np.log(np.maximum(np.finfo(x.dtype).smallest_normal, x))
```
stephanlachnit, line 119: It's probably faster if `np.finfo(x.dtype).smallest_normal` is placed in a separate variable so that it doesn't need to be evaluated every time. You can find my current patch [here](https://salsa.debian.org/science-team/iminuit/-/blob/95b55b5e26a8796f65185cd89167903f6451720d/debian/patches/Fix-safe-log-precision.patch) (includes a pytest adjustment).

stephanlachnit:
> Could you please check whether this fix indeed solves the issue? I cannot run tests on the affected systems.

Unfortunately, it does not. You can find the logs for different architectures [here](https://buildd.debian.org/status/package.php?p=iminuit).

I do have access to these architectures in theory. I can try to look into it again and maybe test it with some verbose output to see what's going on there.

src/iminuit/cost.py
HDembinski: In that case, we cannot select the subnormal that matches the dtype of x, but the current code does not work with numba anyway, so probably that's the solution to go for.

HDembinski: I went for your solution, could you check again? Now Numba works again.

HDembinski:
Really nice to see that Debian is distributing iminuit. When I was younger, I used Debian for many years.

HDembinski:
I am going to merge this, because it is definitely more safe as the old code, since it can handle negative x (which should never happen, but you never know).

If the issue persists, please reopen the issue.

PR #922, opened 2023-08-04
Support Interval type and check compatibility with pydantic

HDembinski:
This adds support for the Interval annotation type that is in the package `annotated_types` and `pydantic`, which was previously missing. It also adds tests for float annotations from `annotated_types` and `pydantic`, to check that they are correctly handled by `util.describe`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,8 @@ test = [
     "boost_histogram",
     "resample",
     "unicodeitplus",
+    "pydantic",
+    "annotated_types",
 ]
 doc = [
     "sphinx<7",
--- a/src/iminuit/typing.py
+++ b/src/iminuit/typing.py
@@ -71,3 +71,13 @@ class Le:
     """Annotation compatible with annotated-types."""
 
     le: float
+
+
+@dataclasses.dataclass
+class Interval:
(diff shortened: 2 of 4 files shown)


HDembinski:
@matthewfeickert I requested a review from you, because Github suggested it. Maybe it knows that you have been dealing with pydantic or Annotated types?

matthewfeickert:
> I requested a review from you, because Github suggested it. Maybe it knows that you have been dealing with pydantic or Annotated types?

Thanks @HDembinski. Perhaps GitHub knows from my browsing that I want to get more into Pydantic, but I must admit that I really have zero experience with it so far and so don't think I can give you any useful review.

If you do want someone with lots of Pydantic experience I would suggest [`adonath`](https://github.com/adonath) (who I think you know?) and gave a great Pydantic tutorial at SciPy 2023.

HDembinski:
@adonath If you have the time, it would be great if you had a look over how iminuit is starting to support restricted float type from `pydantic` and `annotated_types`. The idea is to use them to declare limits on model parameters when specifying the model, instead of setting those limits when calling Minuit.

HDembinski:
I am going to merge now, since it is not a controversal addition to the functionality that was already there.

PR #946, opened 2023-10-11
fix: include debug info on failures

henryiii:
WIP, Adding debug info on failures.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,5 +92,6 @@ else()
 endif()
 target_include_directories(_core PRIVATE extern/root/math/minuit2/inc)
 set_target_properties(_core PROPERTIES VISIBILITY_INLINES_HIDDEN ON)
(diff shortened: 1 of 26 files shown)


henryiii:
The newly detailed error:

```
RuntimeError: Unable to cast Python instance of type <class 'numpy.ndarray'> to C++ type 'double'
```

henryiii:
Ahah, _all_ warnings as errors is not enabled. This NumPy warning apparently has been around for a while:

```
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation.
```

I think NumPy 1.26 is likely causing the breakage.

henryiii:
PS: Why is this only showing up in the coverage job?

henryiii:
Okay, great, let me know if I can help.

PR #950, opened 2023-11-22
Approximate CDF from PDF

The-Ludwig:
- make the description of the approximate CDF more precise 
- provide an example of how to generate an approximate CDF from a given PDF
- comments out 2 lines so the notebook throws no errors (both related to the `jacobi` example)


HDembinski:
@The-Ludwig Let's not touch this branch for now, I am working on a different solution. I also reworked the notebook that you changed, so I will merge this PR at some point. If you want to do something, you could make a new notebook which show-cases the different ways of computing an cdf (approximate or with numerical integration) and showcase the pros and cons. You could explore the bias introduced by the approximate cdf and compare run-times.

PR #977, opened 2024-03-31
Support NumPy 2

henryiii:
A new release of iMinuit needs to be made with pybind11 2.12 in order to support numpy 2. All binaries must be rebuilt with pybind11 2.12+ in order to support NumPy 2.

That should happen automatically, but I took the opportunity to update a few things while I was adding a NumPy 2 test. (which I actually forgot to do, will make draft until I add that!)


henryiii:
VisibleDeprecationWarning has been ~~re~~moved in NumPy 2. Actually, I think it was always supposed to be `numpy.exceptions.VisibleDeprecationWarning`, it just used to be leaked into the main namespace in NumPy 1.

.github/workflows/release.yml
```diff
@@ -37,20 +37,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-        arch: [auto, aarch64, universal2]
+        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
```
henryiii, line 40: MacOS 14 is the new ARM runner. So this builds native wheels on both (the ARM part can now be tested! Except for CPython 3.8, which is wonky)

pyproject.toml
```diff
@@ -126,7 +126,7 @@ no_implicit_optional = false
 
 [tool.cibuildwheel]
 # update skip when numpy wheels become available
-skip = ["*-musllinux_*", "cp31?-win32", "cp31?-manylinux_i686"]
+skip = ["cp31?-manylinux_i686"]
```
henryiii, line 129: NumPy has binaries for these now. But we should run a wheel build to be sure!

henryiii:
```suggestion
skip = ["*-musllinux_*", "cp31?-manylinux_i686"]
```

Oops, not musl.

henryiii:
Wheel build: https://github.com/henryiii/iminuit/actions/runs/8520504460/job/23336798970

.github/workflows/test.yml
```diff
@@ -18,15 +18,19 @@ jobs:
...
             python-version: "3.11"
+            installs: "numpy>=2.0.0rc1"
           - os: macos-latest
             python-version: "3.8"
+            installs: "'numpy==1.21.0'"
           - os: ubuntu-latest
             python-version: "pypy-3.8"
+          - os: ubuntu-latest
+            python-version: "3.12"
+            installs: "'numpy>=2.0.0rc1' 'scipy>=1.13.0rc1'"
```
henryiii, line 33:
```suggestion
            installs: "'numpy>=2.0.0rc1' 'scipy>=1.13.0'"
```

henryiii:
SciPy 1.13 has shipped! One step closer to NumPy 2.0 releasing. `pip install iminuit` will start breaking in Python 3.9+ once NumPy 2 ships if we don’t make a new release. Boost-histogram and awkward have released new compatible versions.

.github/workflows/test.yml
```diff
@@ -41,8 +45,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - run: python -m pip install --upgrade pip wheel
     # python -m pip install .[test] is not used here to test minimum (faster),
-    # the cov workflow runs all tests
-    - run: python -m pip install -v . pytest
+    # the cov workflow runs all tests.
+    - run: python -m pip install -v . pytest ${{ matrix.installs }}
```
HDembinski, line 49: This looks like it can be handled in pyproject.toml with environment markers. https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers

src/iminuit/cost.py
```diff
@@ -1521,7 +1522,7 @@ def __init__(self, n, xe, model, verbose, grad, use_pdf):
                 self._model_xm = self._model_xe[:-1] + 0.5 * dx
         else:
             self._xe_shape = tuple(len(xei) for xei in self.xe)
-            self._model_xe = np.row_stack(
+            self._model_xe = np.vstack(
```
HDembinski, line 1525: Why this replacement?

HDembinski:
> FYI, Python has a visible Deprecation warning. It's called [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning): _Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python._ (Wording was updated in a PEP a few years ago to clarify it is intended to be a deprecation warning)

Ok, I shall use that in the future.

HDembinski:
Thank you for the extensive patch and the pointers. If feasible, I would prefer to install the right versions of dependencies via environment markers declared in pyproject.toml. This should be possible for the numpy version at least. scipy is an optional dependency, I don't know whether these can be declared with environment markers as well. If it is cannot be done easily, then installing the right scipy version in the CI workflow is ok.

src/iminuit/cost.py
henryiii: From the old docs, they've always been the same function:

> `np.row_stack` is an alias for [vstack](https://numpy.org/doc/stable/reference/generated/numpy.vstack.html#numpy.vstack). They are the same function.

And they are getting rid of the `row_stack` name in favor of `vstack`:

https://numpy.org/devdocs/numpy_2_0_migration_guide.html#main-namespace

The current one throws a warning. Ruff's NPY check will also flag this.

.github/workflows/test.yml
henryiii: The goal is to try with the minimum and maximum supported numpy version. Everything from 1.21 to 2.0.0rc1 is supported on ll platforms, this is just a sampling for the tests.

Another way to do it would be to switch from using `pip` to `uv`, then `uv pip install -e . --resolution=lowest-direct` will install your minimum dependencies. I was actually using that locally in testing the patch. And `uv pip install . --prerelease` (regular pip supports this too) will install pre-releases, which would get numpy 2.0.0rc1.

henryiii: (If you do have a dependency that depends on the install environment, environment markers are exactly the way to go, but this is not that, it's just trying a selection of min and max deps for testing)

.github/workflows/test.yml
```diff
@@ -18,15 +18,19 @@ jobs:
...
             python-version: "3.11"
+            installs: "numpy>=2.0.0rc1"
           - os: macos-latest
             python-version: "3.8"
+            installs: "'numpy==1.21.0'"
           - os: ubuntu-latest
             python-version: "pypy-3.8"
+          - os: ubuntu-latest
+            python-version: "3.12"
+            installs: "'numpy>=2.0.0rc1' 'scipy>=1.13.0'"
```
henryiii, line 33:
```suggestion
            installs: "'numpy>=2.0.0rc1' scipy matplotlib"
```

Matplotlib 3.8.4 just came out, which supports numpy 2. Also, there's no need to put a `>=` here unless there are only dev releases, so I removed the `>=1.13` on scipy.

.github/workflows/test.yml
HDembinski: Ok, thanks for clarifying, after I wrote that I also got some doubts whether I understood correctly what you are doing, which you confirm here. Thanks for the hint to uv, that seems pretty cool.

src/iminuit/cost.py
HDembinski: Ok!

HDembinski:
Thank you for the nice patch.

matthewfeickert:
:wave: @HDembinski would it be possible to get a release of some sort (I'm not sure given the [development that has happened since release `v2.25.2`](https://github.com/scikit-hep/iminuit/compare/v2.25.2...a2fe2efe61e0a70e004603c340483a6a822b87dc) what SemVer type it should be) to support NumPy 2?

```console
$ docker run --rm -ti python:3.12 /bin/bash
root@f790da29487c:/# python -m venv venv && . venv/bin/activate
(venv) root@f790da29487c:/# python -m venv venv && . venv/bin/activate
(venv) root@f790da29487c:/# python -m pip --quiet install --upgrade pip wheel
(venv) root@f790da29487c:/# python -m pip install --upgrade iminuit
Collecting iminuit
  Downloading iminuit-2.25.2-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (11 kB)
Collecting numpy>=1.21 (from iminuit)
  Downloading numpy-1.26.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (61 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 61.0/61.0 kB 926.1 kB/s eta 0:00:00
Downloading iminuit-2.25.2-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (423 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 423.9/423.9 kB 6.2 MB/s eta 0:00:00
Downloading numpy-1.26.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (18.0 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 18.0/18.0 MB 42.7 MB/s eta 0:00:00
Installing collected packages: numpy, iminuit
Successfully installed iminuit-2.25.2 numpy-1.26.4
(venv) root@f790da29487c:/# python -m pip install --upgrade --pre numpy
Requirement already satisfied: numpy in /venv/lib/python3.12/site-packages (1.26.4)
Collecting numpy
  Downloading numpy-2.0.0rc2-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (60 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 60.9/60.9 kB 937.6 kB/s eta 0:00:00
Downloading numpy-2.0.0rc2-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (19.0 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 19.0/19.0 MB 52.2 MB/s eta 0:00:00
Installing collected packages: numpy
  Attempting uninstall: numpy
    Found existing installation: numpy 1.26.4
    Uninstalling numpy-1.26.4:
      Successfully uninstalled numpy-1.26.4
Successfully installed numpy-2.0.0rc2
(venv) root@f790da29487c:/# python
Python 3.12.3 (main, Apr 24 2024, 11:17:35) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import iminuit
  File "<stdin>", line 1, in <module>
  File "/venv/lib/python3.12/site-packages/iminuit/__init__.py", line 24, in <module>
    from iminuit.minuit import Minuit
  File "/venv/lib/python3.12/site-packages/iminuit/minuit.py", line 6, in <module>
    from iminuit import util as mutil
  File "/venv/lib/python3.12/site-packages/iminuit/util.py", line 11, in <module>
    from iminuit import _repr_html, _repr_text, _deprecated
  File "/venv/lib/python3.12/site-packages/iminuit/_deprecated.py", line 2, in <module>
    from numpy import VisibleDeprecationWarning
ImportError: cannot import name 'VisibleDeprecationWarning' from 'numpy' (/venv/lib/python3.12/site-packages/numpy/__init__.py)
```

I've started to see this in some environments where other dependencies are testing against the release candidates of NumPy, and it would be great if `iminuit` could have a release that supports them, as these changes are already in the repository.

henryiii:
NumPy 2 will be released mid June. June 16 IIRC. I would highly recommend releasing an iminuit release supporting it before then!

HDembinski:
@henryiii @matthewfeickert Release is in progress.

henryiii:
Fantastic, thanks! I think this means all our key libraries are now ready for NumPy 2.

PR #978, opened 2024-03-31
Use modern Ruff config & Ruff fmt

henryiii:
- **chore: update ruff config**
- **chore: use ruff fmt**


pyproject.toml
```diff
@@ -94,23 +94,16 @@ filterwarnings = [
...
+src = ["src"]
+lint.extend-select = [
     "D", # pydocstyle
 ]
-extend-ignore = [
+lint.ignore = [
     "D212", # multi-line-summary-first-line
 ]
-src = ["src"]
-unfixable = [
```
henryiii, line 106: This is now an "unsafe" fix, and has to be enabled with `--unsafe-fixes`. So it's no longer needed to put in in unfixable.

pyproject.toml
```diff
@@ -94,23 +94,16 @@ filterwarnings = [
 ]
 
 [tool.ruff]
-select = [
-    "E",
-    "F", # flake8
+src = ["src"]
+lint.extend-select = [
```
henryiii, line 98: E and F (with a few useful exclusions that conflict with formatters) are in the default set, so extend-select is better.

pyproject.toml
```diff
@@ -94,23 +94,16 @@ filterwarnings = [
...
     "D", # pydocstyle
 ]
-extend-ignore = [
+lint.ignore = [
```
henryiii, line 101: Conversely, `extend-ignore` was exactly the same as `ignore`, and is deprecated. (and everything lint-specific is supposed to be in a `lint` subsection now since Ruff is also a formatter)

src/iminuit/minimize.py
```diff
@@ -85,7 +85,7 @@ def f(x):
     if bool(jac):
         if jac is True:
             raise ValueError("jac=True is not supported, only jac=callable")
-        assert hasattr(jac, "__call__")
+        assert callable(jac)
```
henryiii, line 88: I tried to enable B, and there were a few other things that needed manual fixing, so I left it off this PR. But this change seemed nice.

PR #995, opened 2024-05-31
ci: add pyodide build

henryiii:
Testing this.


.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
   workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/release.yml
```
HDembinski, line 10: Nice

henryiii: To give credit, I first saw this here: https://github.com/astral-sh/uv/blob/fbf562d5a86a58d857ed31545f3faed9ed837da7/.github/workflows/build-binaries.yml#L16-L20 - it's great for working on this file, saves so much time going to manually trigger workflows. I think having it trigger on pyproject.toml would be a bit much, but this is an easy win.

henryiii: I think everything is fixed upstream, let's see! :)

henryiii:
```
  ______________________ test_NormalConstraint_bad_input_4 _______________________
      def test_NormalConstraint_bad_input_4():
  >       with pytest.raises(ValueError, match="positive definite"):
  E       Failed: DID NOT RAISE <class 'ValueError'>
  /home/runner/work/iminuit/iminuit/tests/test_cost.py:1573: Failed
  ______________________ test_NormalConstraint_bad_input_5 _______________________
      def test_NormalConstraint_bad_input_5():
          n = NormalConstraint(["a", "b"], [1, 2], [[1, 0], [0, 1]])
      
  >       with pytest.raises(ValueError, match="positive definite"):
  E       Failed: DID NOT RAISE <class 'ValueError'>
  /home/runner/work/iminuit/iminuit/tests/test_cost.py:1580: Failed
  ____________________________ test_positive_definite ____________________________
      def test_positive_definite():
          assert util.is_positive_definite([[1, 0], [0, 1]])
  >       assert not util.is_positive_definite([[1, 1], [1, 1]])
  E       assert not True
  E        +  where True = <function is_positive_definite at 0x2898118>([[1, 1], [1, 1]])
  E        +    where <function is_positive_definite at 0x2898118> = util.is_positive_definite
  /home/runner/work/iminuit/iminuit/tests/test_util.py:776: AssertionError
```

HDembinski:
> This is the only failing part, don't know if it's directly pyodide's fault:
> 
> ```
>   ______________________ test_NormalConstraint_bad_input_4 _______________________
>       def test_NormalConstraint_bad_input_4():
>   >       with pytest.raises(ValueError, match="positive definite"):
>   E       Failed: DID NOT RAISE <class 'ValueError'>
>   /home/runner/work/iminuit/iminuit/tests/test_cost.py:1573: Failed
>   ______________________ test_NormalConstraint_bad_input_5 _______________________
>       def test_NormalConstraint_bad_input_5():
>           n = NormalConstraint(["a", "b"], [1, 2], [[1, 0], [0, 1]])
>       
>   >       with pytest.raises(ValueError, match="positive definite"):
>   E       Failed: DID NOT RAISE <class 'ValueError'>
>   /home/runner/work/iminuit/iminuit/tests/test_cost.py:1580: Failed
>   ____________________________ test_positive_definite ____________________________
>       def test_positive_definite():
>           assert util.is_positive_definite([[1, 0], [0, 1]])
>   >       assert not util.is_positive_definite([[1, 1], [1, 1]])
>   E       assert not True
>   E        +  where True = <function is_positive_definite at 0x2898118>([[1, 1], [1, 1]])
>   E        +    where <function is_positive_definite at 0x2898118> = util.is_positive_definite
>   /home/runner/work/iminuit/iminuit/tests/test_util.py:776: AssertionError
> ```

The two errors seem related to `util.is_positive_definite`, which is implemented as follows:
```py
    m = np.atleast_2d(m)
    if np.all(m.T == m):
        # maybe check this first https://en.wikipedia.org/wiki/Diagonally_dominant_matrix
        # and only try cholesky if that fails
        try:
            np.linalg.cholesky(m)
        except np.linalg.LinAlgError:
            return False
        return True
    return False
```
This returns true in those failing cases although it should return false. Apparently, `np.linalg.cholesky(m)` is not raising `LinAlgError` on Pyiodide.

HDembinski:
Perhaps the line `np.linalg.cholesky(m)` is non-blocking in Pyiodide. This code assumes it is blocking, like any other normal Python statement.

henryiii:
I bet NumPy is compiled with exceptions turned off; maybe this function actually raises the exception from the compiled code?

HDembinski:
@henryiii I made a patch which may work around the failing test in pyodide, but I cannot test it because the pyodide build now fails because of some build-system issue. I am closing this PR now, you can reopen it when you find the time to finish this.

PR #999, opened 2024-06-10
ci: try using uv with cibuildwheel

henryiii:
Testing before cibuildwheel release, this should be faster (and not break/fail ;) ).

~~Ignore the validator fail, will work shortly after we release.~~ (pushed to SchemaStore and updated).

Ahh, no uv in manylinux2010, updates for manylinux1 and manylinux2010 discontinued.

Seems to save ~15 seconds to a minute on each job.


.github/workflows/release.yml
```diff
@@ -56,7 +59,9 @@ jobs:
       - if: ${{ matrix.arch == 'aarch64' }}
         uses: docker/setup-qemu-action@v3
 
-      - uses: pypa/cibuildwheel@v2.17.0
+      - uses: yezz123/setup-uv@v4
+
+      - uses: pypa/cibuildwheel@main
```
henryiii, line 64:
```suggestion
      - uses: pypa/cibuildwheel@v2.19
```

PR #1024, opened 2024-08-15
fix: missing dependency on packaging

henryiii:
This is used, but not declared. So many packages depend on this that it's being always pulled in for testing, but in a minimal environment it's missing.

See https://github.com/pyodide/pyodide/pull/4767#issuecomment-2291397326.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Operating System :: Unix",
     "Operating System :: MacOS",
 ]
-dependencies = ["numpy >=1.21"]
+dependencies = ["packaging", "numpy >=1.21"]
 
 [project.urls]
 repository = "http://github.com/scikit-hep/iminuit"


HDembinski:
Oh, I was under the impression that is a stdlib package. In that case, I should replace it, but for the time being, this is the right fix.

henryiii:
I wouldn’t recommend replacing it, it is by far the standard tool for this. But it’s not distributed as part of the standard library because it needs to update at a higher frequency.

HDembinski:
I only use it to parse the version number. That can easily be replaced.

henryiii:
It is the canonical way to parse version numbers, and is a more reliable dependency than numpy, so IMO it's completely fine as a dependency. It's a dependency of pip even.

henryiii:
(Not saying you can't remove it if you really don't want it, just saying I think it's not a net positive to remove it. Same goes for other packages from the CPython core developers, like typing-extensions.)

HDembinski:
It is said in the README that iminuit's only hard dependency is on numpy. That is in its simplicity beautiful and I tend to keep it that way. You make good points, but this is a matter of style.

PR #1029, opened 2024-08-25
Fix ROOT version in docs

HDembinski:


HDembinski:
The implemented "fix" does not work, this needs more research to get the right root version on the CI build machine automatically.

PR #1051, opened 2024-10-09
chore: some minor updates

henryiii:
A few things I noticed that can be updated. Modern version of ruff can auto-detect `src` layout. scikit-build-core 0.10+ can read the min version from the `build-system.requires` table.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["scikit-build-core[pyproject]>=0.5", "pybind11>=2.12"]
+requires = ["scikit-build-core>=0.10", "pybind11>=2.12"]
 build-backend = "scikit_build_core.build"
 
 [project]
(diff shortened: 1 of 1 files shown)


HDembinski:
Thanks. Btw, I am working on a patch to get pyhepmc up and running so that we can add a cp313 wheel there as well. I guess it is the last one missing for the scikit-hep package?

henryiii:
I don't think it's in the scikit-hep metapackage.

HDembinski:
Why not?

henryiii:
No idea. I think the goal is to have a somewhat basic subset of packages that tend to be used together. If it fits that, then it could be added. The larger gets, the more unwieldy it gets with potential version incompatibilities, so it should only be added if it’s something that people will generally want to have available with a basic install.

I wonder if we should start defining related subsets of packages as extras…

HDembinski:
Putting it in "extra" oder "all" makes sense. It is an important package, though, HEPMC3 format is a standard serialization format for event generators.

Issue #771, opened 2022-07-28
Minuit.interactive outside of Jupyter notebooks


HDembinski:
Minuit.interactive can detect whether it is run inside a notebook and only then generate the UI with ipywidgets. Outside of a notebook, it could generate the UI with PyQt. This would allow one to call Minuit.interactive for debugging when some fit in a long-running script fails that does a lot of fits at once.

HDembinski:
Hi Carlos, it is not possible at the moment to use the interactive mode outside of a notebook, since it is based on the ipywidgets library at the moment.

This issue exists as a reminder to implement an alternative UI based on PyQt that open in a separate window, but this does not exist yet.

adryyan:
Hi Hans, I have tried getting this to work since this would be really helpful in my projects. I have an implementation for this in my fork that mostly works using PyQt6. When testing my implementation with the examples in the Interactive fits tutorial all of them work except when `CostSum` is used. This seems to be because the visualize method of `CostSum` calls `plt.sca(ax[i])`, which fails for a figure managed by `FigureCanvasQTAgg`. As I understand it, that some `pyplot` functions work with an Event Loop not managed by matplotlib is more accident than intended and when using `FigureCanvasQTAgg` one should really use the axes methods for plotting and avoid `pyplot`. So for this feature to work properly, I think all visualize methods would need to accept a matplotlib figure as a keyword argument and only use `pyplot` if None was provided. So before I continue to spend time on this I thought I'd ask if this would be acceptable or not wanted or if you see another solution. Thanks for all the work on this great package.

HDembinski:
Thank you for your effort, this sounds all sensible, why don't you make a PR then I can have a look?

I would like to avoid breaking the API, perhaps there is a workaround.

Otherwise, I am also ok with deprecating the current visualization API and make one that works on figures. Passing the objects around explicitly instead of using global variables is a better design for a library anyway.

HDembinski:
That's now implemented.

PR #1056, opened 2024-11-01
Minuit.interactive outside of Jupyter notebooks

adryyan:
Implementation of a PyQt6 widget for `minuit.interactive` (Issue #771 ). In order to make the plotting work properly, I changed the API for the `visualize` methods of the build-in cost functions so that they optionally accept matplotlib `Axes` objects (and for `CostSum`s a `Figure`).  If these arguments don't receive a value, `pyplot` is used. This should keep the disruption of existing code as minimal as I could manage.


src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+
+        def on_min_changed(self):
+            tmin = self.tmin.value()
+            if tmin >= self.vmax:
+                self.tmin.blockSignals(True)
+                self.tmin.setValue(self.vmin)
+                self.tmin.blockSignals(False)
+                return
+            self.vmin = tmin
+            self.slider.blockSignals(True)
```
HDembinski, line 121: Since you use signal blocking repeatedly, it is worthwhile to add a context manager for this. It makes the code more readable, and in case of exceptions, signals are still unblocked.

```py
from contextlib import contextmanager

@contextmanger
def SignalBlocker(obj):
     obj.blockSignals(True)
     yield
     obj.blockSignals(False)


# usage

with SignalBlock(obj):
     # do stuff where signals of obj are blocked

File: src/iminuit/minuit.py [hunk 2]
```diff
@@ -2314,13 +2314,19 @@ def interactive(
         self,
         plot: Callable = None,
         raise_on_exception=False,
+        qt_exec=True,
```
HDembinski, line 2317: I don't like this option. I suspect there is a way to find out whether a QApplication is already running, and only start the QApplication and its event loop when it does not already exist. I would prefer that over this option, which is rather technical.

src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+            plot_layout = QtWidgets.QVBoxLayout(plot_group)
+            # Use pyplot here to allow users to use pyplot in the plot
+            # function (not recommended / unstable)
+            self.fig, ax = plt.subplots()
+            self.canvas = FigureCanvasQTAgg(self.fig)
+            plot_layout.addWidget(self.canvas)
+            plot_layout.addStretch()
+            interactive_layout.addWidget(plot_group, 0, 0, 2, 1)
+            try:
+                plot(minuit.values, fig=self.fig)
```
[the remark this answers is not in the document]
HDembinski: You can set the current axes for drawing with `pyplot.sca` and the current figure with `pyplot.scf`.

src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+            self.setFont(font)
+            centralwidget = QtWidgets.QWidget(parent=self)
+            self.setCentralWidget(centralwidget)
+            central_layout = QtWidgets.QVBoxLayout(centralwidget)
+            tab = QtWidgets.QTabWidget(parent=centralwidget)
+            interactive_tab = QtWidgets.QWidget()
+            tab.addTab(interactive_tab, "Interactive")
+            results_tab = QtWidgets.QWidget()
+            tab.addTab(results_tab, "Results")
+            central_layout.addWidget(tab)
```
HDembinski, line 211: You don't need to make a mainwindow here, because you don't use menu bars, a QWidget is enough, I believe. I also don't like that you put the result and the controls onto different tabs. I would like to have them side by side like in the ipywidget.

src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+            button_group = QtWidgets.QGroupBox("", parent=interactive_tab)
+            size_policy = QtWidgets.QSizePolicy(
+                QtWidgets.QSizePolicy.Policy.Expanding,
+                QtWidgets.QSizePolicy.Policy.Fixed,
+            )
+            button_group.setSizePolicy(size_policy)
+            button_layout = QtWidgets.QHBoxLayout(button_group)
+            self.fit_button = QtWidgets.QPushButton("Fit", parent=button_group)
+            self.fit_button.setStyleSheet("background-color: #2196F3; color: white")
+            self.fit_button.clicked.connect(partial(self.do_fit, plot=True))
```
HDembinski, line 250: A lambda is as terse and faster to execute for the Python interpreter.

HDembinski: It is better to use the basic features of the language instead of library functions, if those do not add anything essential.

src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+    sign = -1 if x < 0 else 1
+    if abs(x) == np.inf:
+        return sign * sys.float_info.max
+    return x
+
+
+def _guess_initial_step(val: float, vmin: float, vmax: float) -> float:
+    if np.isfinite(vmin) and np.isfinite(vmax):
+        return 1e-2 * (vmax - vmin)
+    return 1e-2
```
HDembinski, line 413: This look like they were copied from elsewhere. Import them from the ipywidget or from utils, wherever they came from.

HDembinski:
@adryyan I reviewed this PR, but you also made another where you don't change the plotting API. I prefer the other one, as expressed here.

I think we need a few iterations, but I generally appreciate your effort very much. This is a significant contribution.

adryyan:
@HDembinski Thank you for your comments; I will go through them in the next weeks.

src/iminuit/qtwidget.py
adryyan: Makes sense; I changed it in [9485f85](https://github.com/scikit-hep/iminuit/pull/1056/commits/9485f85c203db9d276942ecd88a8a1ceff6295df)

src/iminuit/qtwidget.py
adryyan: agreed [e20f357](https://github.com/scikit-hep/iminuit/pull/1056/commits/e20f357b5734260ac69a74a0e50287850bd44d19)

src/iminuit/qtwidget.py
adryyan: They are from ipywidget, but I don't want to import them from there, since then I would also be importing the ipywidgets package or is there a way to prevent that I am not aware of? I could move these functions into utils and both ipywidget and qtwidget then import them from there?

src/iminuit/minuit.py
adryyan: I removed it, but not in the tests yet. I'll have to figure out how I can make the tests work later.

src/iminuit/qtwidget.py
adryyan: The results are now below the plot. When the fit only has few parameters, it looks kinda stupid though. So if you can think of a better layout, let me know.

src/iminuit/qtwidget.py
HDembinski: Yes, please move them to utils then.

src/iminuit/qtwidget.py
```diff
@@ -0,0 +1,413 @@
...
+
+    if qt_exec:
+        app = QtWidgets.QApplication.instance()
+        if app is None:
+            app = QtWidgets.QApplication([])
+        main_window = MainWindow()
+        main_window.show()
+        app.exec()
+    else:
+        return MainWindow()
```
HDembinski: I haven't tested that, but it looks like this code can be made to work without qt_exec.
```suggestion
    app = QtWidgets.QApplication.instance()
    run_event_loop = False
    if app is None:
         app = QtWidgets.QApplication([])
         run_event_loop = True

    main_window = MainWindow()
    main_window.show()

    if run_event_loop:
        app.exec() # this blocks the main thread
    return main_window
```

That should run the event loop if it does not exist when we call make_widget, but not if the event loop is already running (such as in the tests).

HDembinski: If this does not work, then let's keep the original solution, but rename `qt_exec` to something more sensible, like `start_event_loop = True`.

In the end, this parameter solely exists for the unit tests, right? That was not immediately clear when I read the code.

tests/test_draw.py
```diff
@@ -161,40 +162,138 @@ def assert_call(self):
     plot = Plot()
 
     m = Minuit(cost, 1, 1)
+
+    with patch("IPython.get_ipython") as mock_get_ipython:
+        mock_shell = MagicMock()
+        mock_shell.__class__.__name__ = "ZMQInteractiveShell"
+        mock_shell.config = {"IPKernelApp": True}
+        mock_get_ipython.return_value = mock_shell
```
HDembinski, line 170: This is nice, can you turn that into a pytest fixture?

tests/test_without_ipywidgets.py
```diff
@@ -11,8 +12,14 @@ def test_interactive():
 
     cost = LeastSquares([1.1, 2.2], [3.3, 4.4], 1, lambda x, a: a * x)
 
-    iminuit.Minuit(cost, 1).interactive()
+    with patch("IPython.get_ipython") as mock_get_ipython:
+        mock_shell = MagicMock()
+        mock_shell.__class__.__name__ = "ZMQInteractiveShell"
+        mock_shell.config = {"IPKernelApp": True}
+        mock_get_ipython.return_value = mock_shell
```
HDembinski, line 19: If you turn this code into a fixture like I suggested in test_draw, we can reuse it here without repeating the code.

tests/test_without_pyqt6.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_interactive(qtbot):
+    pytest.importorskip("matplotlib")
+    import iminuit
+
+    cost = LeastSquares([1.1, 2.2], [3.3, 4.4], 1, lambda x, a: a * x)
+    mw = iminuit.Minuit(cost, 1).interactive(qt_exec=False)
+    qtbot.addWidget(mw)
+    mw.close()
+    mw.deleteLater()
```
HDembinski, line 16: This can be removed. We only want to test that our custom ModuleNotFoundError is raised here.

tests/test_without_pyqt6.py
```diff
@@ -0,0 +1,23 @@
...
+
+    with hide_modules("PyQt6", reload="iminuit.qtwidget"):
+        with pytest.raises(ModuleNotFoundError, match="Please install"):
+            mw = iminuit.Minuit(cost, 1).interactive(qt_exec=False)
+            qtbot.addWidget(mw)
+            mw.close()
+            mw.deleteLater()
```
HDembinski, line 23: This is enough:
```suggestion
            iminuit.Minuit(cost, 1).interactive(qt_exec=False)
```
You don't have to explicitly delete qwidgets, Python's memory management takes care of that. But in this case, the widget is not even created.

src/iminuit/qtwidget.py
HDembinski: Please use the same layout like the ipywidget. Use a QHBoxLayout to put the plot left of the controls. You can nest layouts.

src/iminuit/qtwidget.py
adryyan: https://github.com/scikit-hep/iminuit/blob/2813c5131db0e6fe1024ba39d3725937fdc9bb31/src/iminuit/qtwidget.py#L381-L393
My solution would now be this. My idea was, that if no event loop is running, the user probably wants the app to be started by the interactive call and block further code execution in order to debug some fit. To enable multiple interactive calls in one script, I set a name and check if the existing instance has that name. 
I additionally wanted a way for the interactive call to return the widget without starting an event loop for two reasons: One is the tests and the more important second reason is, that one could easily include the widget in a another application.
Also, by using the qtbot fixture in the tests, the name of this qtbot QApplication instance should be different and we should get the widget returned.

tests/test_draw.py
adryyan: I did that now and I put it into conftest.py so that both test_draw.py and test_without_ipywidgets.py can use it.

tests/test_without_pyqt6.py
adryyan: I removed it.

tests/test_without_pyqt6.py
adryyan: In the earlier versions I had the problem that during the tests sometimes the event loop would be started and sometimes the created widgets would interfere with another for some reasons, so I was being extra careful. But that is not happening anymore now and probably got fixed by one of the changes we made. So I removed all that stuff.

src/iminuit/qtwidget.py
adryyan: They are in util now, but I renamed them, because there was already one with the name _guess_initial_step in there. I added a _widget in front of the names.
https://github.com/scikit-hep/iminuit/blob/a50b5e52cff4a492f1a24ebf1c65e38ac60c4f34/src/iminuit/util.py#L1689-L1699

src/iminuit/qtwidget.py
adryyan: I don't understand what you mean exactly: the controls are still to the right of the plot:
Screenshot From 2024-12-11 23-34-39
I am not sure about the layout regarding the results box below, which is not part of the ipywidget, right? So for that I can't use the same layout as the ipywidget. Of course instead of a grid I can use nested vertical and horizontal layouts, but I don't understand what exactly you want changed.

adryyan:
I do not understand this test that was and is still in test_draw.py:
https://github.com/scikit-hep/iminuit/blob/a50b5e52cff4a492f1a24ebf1c65e38ac60c4f34/tests/test_draw.py#L221-L224
Even the comment says that this should raise an exception and I also would expect it to raise an exception, but the test completes without failure even though there is no `with pytest.raises(ValueError):`. This is why I commented this test out for the PyQt6 version.

src/iminuit/qtwidget.py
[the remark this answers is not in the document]
adryyan: > You can set the current axes for drawing with `pyplot.sca`. You can get the figure associated to that axes, too, if necessary.

As mentioned before this is not so simple for a figure that was passed to `FigureCanvasQtAgg` and [using pyplot in such a case is not recommended](https://github.com/matplotlib/matplotlib/issues/19380).

I got it to work after trying a bunch of stuff by manually setting the manager of the canvas:
https://github.com/scikit-hep/iminuit/blob/c54c2d9edb6513e286b43aa346864815f68414b3/src/iminuit/qtwidget.py#L208-L212
This seems somewhat sketchy though. The plotting updates are also considerably slower now, but that might just be me doing stuff inefficiently. But I can understand that this is the preferred solution.

src/iminuit/minuit.py
HDembinski: As mentioned in some another comment, the more I think about it the more ok I am with this, if the user never sees this.

src/iminuit/qtwidget.py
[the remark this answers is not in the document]
HDembinski: Ah thanks for the pointer, I did not know that. Nice workaround!

src/iminuit/qtwidget.py
HDembinski: Ah ok, now I understand what you mean by fit results. I have never tried your code!

In the ipywidget, I don't show the fit result, because it is largely redundant, the parameter values are shown twice and whether the fit succeeded is shown with the text in the top right. In the interactive fit, showing Hesse errors is not very useful. I suggest to remove the block with the fit result.

src/iminuit/qtwidget.py
HDembinski: The widget prefix for guess_initial_step is fine, but please keep the simpler name _make_finite, because that function is general and has no tight link to widgets.

src/iminuit/qtwidget.py
HDembinski: Ok, I see. Using the application name to change behavior here is very brittle, let's not do that. In that case, I rather accept your original code with the keyword qt_exec, but please replace it with "run_event_loop=True", so the default is to run the event loop.

HDembinski:
> I do not understand this test that was and is still in test_draw.py:
> 
> https://github.com/scikit-hep/iminuit/blob/a50b5e52cff4a492f1a24ebf1c65e38ac60c4f34/tests/test_draw.py#L221-L224
> 
> 
> Even the comment says that this should raise an exception and I also would expect it to raise an exception, but the test completes without failure even though there is no `with pytest.raises(ValueError):`. This is why I commented this test out for the PyQt6 version.

This test is to make sure that the iminuit code around the code in class Plot catches an exception raised by class Plot, because if it did not, the interactive plot would stop functioning. Therefore you don't need `with pytest.raises(ValueError):` here.

src/iminuit/qtwidget.py
adryyan: Ok, I removed that box. Are you okay with the way I organized the parameters? I did not want the window to get super wide, but still have a decent size for the control elements. If there are a lot of parameters, the parameter box will become scrollable.

src/iminuit/qtwidget.py
adryyan: Ok, I changed it back to _make_finite.

src/iminuit/qtwidget.py
adryyan: Ok, I added the run_event_loop argument, but did not add it to the docstring.

adryyan:
I still don't understand: the associated ipywidget with that `fit_button.click()` is this:
https://github.com/scikit-hep/iminuit/blob/811b27b950fda0993d1e716fcdd8d1cd0ecb395e/tests/test_draw.py#L209-L224
In the interactive call `raise_on_exception` is set to true, so the ipywidget shouldn't catch the exception? Or am I missing something?

src/iminuit/qtwidget.py
HDembinski: Yeah, that is great, well done. If it is possible I will adapt the ipywidget to look more like this.

HDembinski: I also like the colors for the fit and reset buttons

HDembinski: Here is another idea as a compromise. The "Minuit.fmin" object is also displayable, it will render as the part above the parameter block. It might be a good idea to show that under the plot and the remove the "success" field on top right in the plot. If we decide that this looks good here, I would also change the ipywidget to follow that design.

src/iminuit/minuit.py
```diff
@@ -2347,10 +2352,29 @@ def interactive(
...
+
+            if (
+                get_ipython().__class__.__name__ == "ZMQInteractiveShell"
+                and "IPKernelApp" in get_ipython().config
+            ):
+                is_jupyter = True
+            else:
+                is_jupyter = False
+        except Exception:
+            is_jupyter = False
```
HDembinski, line 2366: This needs to be factored out into an utility function called "is_jupyter", please put in utils.

src/iminuit/qtwidget.py
HDembinski: Ok, wait a minute, why do we have to expose "run_event_loop" to the user at all in Minuit.interactive? We don't need that for testing, I believe? I would like to avoid exposing run_event_loop to the top level user interface. Not documenting the keyword is not a good option, undocumented keywords exposed by an API are bad.

Can you make your use-case work of including the widget in another application without adding that to Minuit.interactive? Can't you call make_widget directly in your other application? It is ok that make_widget has that option, because that is a semi-private API that the user is not expected to know.

src/iminuit/minuit.py
HDembinski: When you do that, can you please also put a reference for this detection magic? I wonder how brittle this is. The ZMQ prefix looks suspiciously like an implementation detail.

HDembinski: I did some research. ChatGPT gives your method as the first of several options, as does Stackoverflow. Digging a bit deeper into the SO answers, I found this technique, which I like slightly better:
https://github.com/tqdm/tqdm/blob/master/tqdm/autonotebook.py

```
def is_jupyter_notebook():
    try:
        from IPython import get_ipython
        if 'IPKernelApp' not in get_ipython().config:  # pragma: no cover
            return False
    except ImportError:
        return False
    except AttributeError:
        return False
    return True
```

This also seems to depend on some implementation details, but those seem less likely to change.

HDembinski:
> In the interactive call `raise_on_exception` is set to true, so the ipywidget shouldn't catch the exception? Or am I missing something?

Ok, I will check this more thoroughly to give you a proper answer.

src/iminuit/minuit.py
HDembinski: Here is another option from IPython issues, although not confirmed by the IPython devs https://github.com/ipython/ipython/issues/9732

```py
def is_jupyter_notebook():
    try:
         from IPython import get_ipython
         ip = get_ipython()
         return ip.has_trait('kernel')
    except ImportError:
         return False
    return False
```

adryyan: I went with the second one for now and moved it into util.

src/iminuit/qtwidget.py
adryyan: Ah okay, I understand now. Yes, that will work for me. I changed the tests in test_draw.py so that they use make_widget directly.

src/iminuit/qtwidget.py
adryyan: Ok, there is some empty space there, but I like having fmin displayed there:
Screenshot From 2024-12-13 20-32-39

HDembinski: Me, too, looks great and is much more intuitive.

src/iminuit/qtwidget.py
HDembinski: Cool, thanks.

src/iminuit/util.py
```diff
@@ -1684,3 +1684,29 @@ def is_positive_definite(m: ArrayLike) -> bool:
...
+def is_jupyter() -> bool:
+    try:
+        from IPython import get_ipython
+
+        ip = get_ipython()
+        return ip.has_trait("kernel")
+    except ImportError:
+        return False
+    except AttributeError:
+        return False
```
HDembinski, line 1698: I think that is not necessary, the object returned by get_ipython always has the "has_trait" method.

HDembinski:
@adryyan We are coming to the finish line. Thank you for all the hard work! This is a significant addition to iminuit, and I praise your improvements to the fitting GUI and your nice unit tests. I will mimic the new GUI in a separate PR at some later point in the ipywidget widget.

The last step is to make sure that all tests pass, then we can merge this PR.

HDembinski:
This one looks a bit tricky https://github.com/scikit-hep/iminuit/actions/runs/12353782099/job/34473645329?pr=1056#step:9:41
`INTERNALERROR> ImportError: libEGL.so.1: cannot open shared object file: No such file or directory`

If you cannot figure it out, I will handle that one.

src/iminuit/util.py
adryyan: I would have thought so as well, but when calling get_ipython() not in a jupyter notebook / ipython session, it just returns None and that of course has no attribute has_trait.

adryyan:
Thank you for all your input, I learned some new things and I feel like the code is much cleaner now.

I don't really have any experience with nox, but the problems look to me like there are issues with the installed packages and importskips. I moved all tests regarding the qtwidget into one file and put the importskip of pyqt6 before the test functions, because of the qtbot (pytest-qt). I did the same with the ipywidget tests, because the mock_ipython fixture fails if there is no IPython. But I think the actions would also need to be updated to install the necessary packages, so that the qtwidget can be tested (PyQt6, pytest-qt), but I am not sure how to do that.

src/iminuit/util.py
HDembinski: Ok, then please add a comment since this is surprising.

adryyan: https://github.com/scikit-hep/iminuit/blob/3167f0f595ec4f39f2981110a49570e91fc06e38/src/iminuit/util.py#L1697-L1699
I added this comment.

HDembinski:
Fixing the CI issue turns out to be difficult. I am getting segfaults when the PyQt code is executed, presumably because there is a mismatch between the versions of libraries that Qt links to at run-time and those which the PyQt Python package was build against. I found this nice install qt action, which should avoid exactly that issue, but it does not work.

One way to fix this is to install all the required packages with conda/mamba, then all the external libs should be consistent with the libs used to build the python package.

Another idea is to port this all to PySide, which is now officially supported by Qt. It is always annoying when the error only appears on CI, that makes debugging difficult. Although it is not clear whether that would avoid the issue.

adryyan:
Changing from PyQt6 to PySide6 is quite easy, so it doesn't hurt to try.

adryyan:
I am not very familiar with GitHub actions and nox; so changing to conda in coverage.yml would entail installing conda from for example miniforge and then changing the nox options for cov in the noxfile?
I can't start the workflows, so could you please approve the pending ones to see if changing to PySide already fixed the problem?

HDembinski:
Very impressive, thanks for tracking this down.

Issue #1058, opened 2024-12-03
`NormalConstraint` is missing from API reference


lvarriano:
For some reason, the `NormalConstraint` cost function is missing from the API reference (https://scikit-hep.org/iminuit/reference.html#).

HDembinski:
Huh, that IS strange! Thanks for noticing, I will try to fix...

PR #1059, opened 2024-12-04
fix NormalConstraint not appearing in docs

lvarriano:
Found the missing line needed to address #1058


HDembinski:
Ah, that is perfect. Your fix is most appreciated.

PR #1060, opened 2024-12-04
Add ability to pass options to SciPy minimizers

lvarriano:
The various SciPy minimizers have different options that can be adjusted to change and improve performance. Presently, the iminuit API does not allow passing options to these minimizers, which limits their usefulness for some minimization problems (specifically mine). I added a parameter to minuit.scipy() to allow passing options to scipy.optimize.minimize(). It keeps the existing default options that were being passed and allows for these defaults to be overwritten by the user.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -987,7 +987,7 @@ class Minuit:
         hess: Any = None,
         hessp: Any = None,
         constraints: Iterable = None,
-        options: Dict[str, ...] = {},
+        options: Dict[str, Any] = {},
     ) -> "Minuit":
         """
         Minimize with SciPy algorithms.
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -1034,10 +1034,10 @@ class Minuit:
         the tolerance :attr:`tol` has no effect on SciPy minimizers.
 
         You can specify convergence tolerance and other options for the SciPy minimizers
-        through the `options` parameter. Note that providing the SciPy options 
-        `"maxiter"`, `"maxfev"`, and/or `"maxfun"` (depending on the minimizer) takes 
-        precedence over providing a value for `ncall`. If you want to explicitly control 
-        the number of iterations or function evaluations for a particular SciPy minimizer, 
+        through the `options` parameter. Note that providing the SciPy options
+        `"maxiter"`, `"maxfev"`, and/or `"maxfun"` (depending on the minimizer) takes
+        precedence over providing a value for `ncall`. If you want to explicitly control
+        the number of iterations or function evaluations for a particular SciPy minimizer,
         you should provide values for all of its relevant options.
         """
         try:
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -987,6 +987,7 @@ class Minuit:
         hess: Any = None,
         hessp: Any = None,
         constraints: Iterable = None,
+        options: Dict[str, ...] = {},
     ) -> "Minuit":
         """
         Minimize with SciPy algorithms.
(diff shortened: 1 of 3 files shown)


src/iminuit/minuit.py
```diff
@@ -1027,6 +1032,9 @@ def scipy(
         criterion is evaluated only after the original algorithm already stopped. This
         means that usually SciPy minimizers will use more iterations than Migrad and
         the tolerance :attr:`tol` has no effect on SciPy minimizers.
+
+        You can specify convergance tolerance and other options for the SciPy minimizers
```
HDembinski, line 1036:
```suggestion
        You can specify convergence tolerance and other options for the SciPy minimizers
```

src/iminuit/minuit.py
```diff
@@ -1224,18 +1235,29 @@ def __call__(self, par, v):
...
+                del scipy_options["maxiter"]
 
         if method in ("L-BFGS-B", "TNC"):
-            options["maxfun"] = ncall
-            del options["maxiter"]
+            if "maxfun" not in options:
+                scipy_options["maxfun"] = ncall
+
+            if "maxiter" not in options:
+                del scipy_options["maxiter"]
```
HDembinski, line 1260: Are these changes an improvement? I think it makes more sense to uniformly set a stopping criterion with ncall. I don't see how overriding this feature with option is helpful.

src/iminuit/minuit.py
```diff
@@ -987,6 +987,7 @@ def scipy(
         hess: Any = None,
         hessp: Any = None,
         constraints: Iterable = None,
+        options: Dict[str, Any] = {},
```
HDembinski, line 990:
```suggestion
        options: Optional[Dict[str, Any]] = None,
```

tests/test_scipy.py
```diff
@@ -231,6 +231,12 @@ def test_bad_constraint():
         m.scipy(constraints=[{}])
 
 
+def test_bad_options():
+    m = Minuit(fcn, a=1, b=2)
+    with pytest.raises(ValueError):
+        m.scipy(options=[])
+
+
```
HDembinski, line 239: Would be nice to also have a test cases that shows the benefit of setting options.

HDembinski:
Thanks, I am happy to include this patch, after some refinement.

src/iminuit/minuit.py
```diff
@@ -1224,18 +1235,29 @@ def __call__(self, par, v):
             else:
                 method = "BFGS"
 
+        # copy to avoid mutable problems
+        scipy_options = dict(options)
```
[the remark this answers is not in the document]
lvarriano: Thanks very much for showing me a better way to do this!

src/iminuit/minuit.py
HDembinski: Ok, after thinking more about it, I take this back. My first impression was that "maxiter", "maxfev", "maxfun" in options should not override ncall, but your solution is ok: if these options are explicitly set, then they override ncall, but if they are not set, ncall is used accordingly. options is for experts, so when you use it you are expected to know what you are doing, and it would be more surprising if setting maxiter etc wouldn't have the desired effect.

src/iminuit/minuit.py
```diff
@@ -1043,6 +1051,9 @@ def scipy(
         if ncall is None:
             ncall = self._migrad_maxcall()
 
+        if type(options) is not dict:
+            raise ValueError("options must be a dictionary")
+
```
HDembinski, line 1056: I am not sure whether this is really necessary. One usually does not check types dynamically in Python, unless the type is used to determine what the code does.

HDembinski: Type checking is a job for mypy and the linter in your code editor.

HDembinski:
@lvarriano Are you still working on that test? Looks like this is almost done and it would be nice to merge soon.

The test should be something simple, but you would certainly know better. You said that by setting certain options you were able to improve performance. I suggest to add a toy example where this can be seen. Performance would be measured in numbers of function calls until convergence.

lvarriano:
Sorry for the delay! We've added a simple test to show the precision of the result improves when the SciPy tolerances are adjusted. This is a stand-in for a problem we're having performing extended unbinned analyses with low amounts of data, which is typical for our experiment.

HDembinski:
Thanks, that is a good demonstration. Without options you cannot manipulate the stopping criterion. Minuit has its own tolerance, but it cannot be easily mapped to the multiple tolerances that scipy offers.

PR #1100, opened 2025-08-08
Add free-threading wheels (missing!)

henryiii:
A little modernization, but mostly adding the missing 313t classifier to build free-threading wheels. If #1099 is merged first, this should be rebased, otherwise that one should be rebased. Also 314t should be added.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-13, macos-latest, ubuntu-24.04-arm, windows-11-arm]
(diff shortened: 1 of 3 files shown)


HDembinski:
@henryiii #1099 was merged, and I rebased this, but building the cp313t wheels fails with a strange linker error.

HDembinski:
The failure of Coverage / coverage is unrelated to this patch. An update in one of the optional dependencies broke the tests. I cannot reproduce this on my local computer, so difficult to track this down.

henryiii:
I believe this is a warning (treated as an error) that the runtime value could be larger than the maximum allocatable size. GCC can't know the size of `num_shards` at compile time, so it assumes it could be too large. There must be some warning enabled here that we don't have enabled in pybind11. I'll see if we can fix it there.

PR #1126, opened 2026-02-24
Feature/custom pdf integrator

Idg772:
# Add `use_pdf="custom"` with `integrator=` parameter for `BinnedNLL` and `ExtendedBinnedNLL`

Closes #1115

Note: This is my first open source contribution so any notes, feedback or suggestions would be very welcome :)

## Motivation

When fitting a binned histogram without an analytical CDF, iminuit currently offers two options via `use_pdf`:

| `use_pdf` | Method | Accuracy | Speed | Dimensions |
|-----------|--------|----------|-------|------------|
| `"approximate"` | PDF at bin centre × bin width | Low for coarse bins | Fast | Any |
| `"numerical"` | `scipy.integrate.quad` per bin | High | Slow | 1D only |

There is no supported way to use a custom integration rule e.g. Simpson's rule that sits between these two in terms of accuracy and speed, and works for multi-dimensional histograms.

### Implementation (`src/iminuit/cost.py`)

- Added `use_pdf="custom"` as a valid option to `BinnedNLL` and `ExtendedBinnedNLL`.
- Added an `integrator` keyword argument to both classes. It is required when `use_pdf="custom"` and raises `ValueError` at construction time if omitted.
- Added `BinnedCostWithModel._make_pred_custom()`, which wraps the user's integrator into the internal `_pred_impl` callable. It handles both 1D and N-D histograms:
  - **1D**: calls `integrator(pdf, a, b)` with scalar bin edges.
  - **N-D**: iterates over all bins via `itertools.product` and calls `integrator(pdf, a, b)` where `a` and `b` are 1-D arrays of length N (corners of the hyper-rectangle).

The integrator interface is:

```python
integrator(pdf, a, b) -> float
```

where `pdf` is the model PDF with parameters already bound, and `a`, `b` are the bin edges. Example with Simpson's rule:

```python
from scipy.integrate import simpson

def simpson_integrator(pdf, a, b):
    x = np.linspace(a, b, 11)
    return simpson(pdf(x), x=x)

cost = ExtendedBinnedNLL(n, xe, scaled_pdf, use_pdf="custom", integrator=simpson_integrator)
```

I chose this implementation over the suggested `cost.register_pdf_integrator` as this allows errors to happen when the cost is defined rather than at minimisation time.

### Tests (`tests/test_cost.py`)

Six new tests covering:
- 1D `BinnedNLL` with custom (Simpson) integrator — checks predictions match the analytical CDF to `rtol=1e-5`
- 2D `BinnedNLL` with custom integrator — checks the N-D code path where `a`/`b` are length-2 arrays
- `BinnedNLL` error when `use_pdf="custom"` is given without `integrator=`
- Same three tests for `ExtendedBinnedNLL`

### Documentation (`doc/notebooks/custom_integrator.ipynb`)

New tutorial notebook added to `doc/tutorials.rst` covering:
- The integrator interface and when to use `"custom"` vs the built-in modes
- 1D example with accuracy comparison (approximate vs quad vs Simpson) and timing benchmark
- `ExtendedBinnedNLL` example with a scaled PDF
- 2D example showing how to write a double-Simpson integrator
- Reference section with alternative quadrature rules (Gauss-Legendre, adaptive quad)

## Notes for reviewers

- `use_pdf="custom"` without `integrator=` raises `ValueError` at construction time (not at fit time), so there is no invalid-state window.
- `_model_xm` and `_model_dx` (used by `"approximate"`) are not allocated when `use_pdf="custom"`.
- The existing `"approximate"`, `"numerical"`, and `""` (CDF) modes are completely unaffected.
- The `grad=` keyword remains incompatible with all `use_pdf` modes (existing validation unchanged).


HDembinski:
I am not quite sure whether this functionality should sit in the cost functions.

The design idea was to give you a quick way to use a model_pdf in simple cases, when you don't have the model_cdf available. The two simple cases are those that are implemented. For everything more complicated than that, users can build their own model_cdf from the model_pdf, and there they can use more complicated integration schemes. This functionality doesn't have to be in the cost functions.

For the N-dimensional case, for example, implementing the integration as a dimension-wise line integration as is done here is inefficient and potentially can lead to errors if the integration routine cannot handle noisy inputs. The sub-integrals generally provide accuracy that's below the full floating point accuracy.

So we can't advocate that as the library solution. We could allow the user to pass a general N-dimensional integrator, but if you have that, it should also be easy to write a small function that implements the model_cdf by internally integrating the model_pdf.

PR #1129, opened 2026-04-12
ci: bump setup-uv to maintained tag scheme

henryiii:
The old vX tags have been dropped to (force) (usually) better security practices. Dependabot will not update, however, leaving this v7 tag forever. Manually updating now.

See https://github.com/astral-sh/setup-uv/issues/830


eduardo-rodrigues:
Hi @henryiii, this seems like a no-brainer and I reckon we can safely merge once the CI is done. I assume Hans is fine with this.

HDembinski:
Yes, I would have been fine with this :)

eduardo-rodrigues:
Let's then merge since the failing test is unrelated and should be addressed elsewhere.

Issue #1132, opened 2026-06-11
Claude Fable review for iminut


henryiii:
:robot: _AI text below_ :robot:

All spot-checks confirmed the agents' top findings. Here is the consolidated review.

# iminuit project review

Five areas reviewed (Python API, cost functions, utilities, C++ bindings, build/CI). The headline: **one wrong-math bug in a fitting hot path, several state-corruption/crash bugs, and a release workflow that will fail on its next run.** I verified the high-severity items directly in the source; several were also reproduced empirically by the review agents (noted below).

## High-severity bugs

**1. `BinnedNLL._grad` returns a wrong gradient with a mask + ≥2 parameters** — `src/iminuit/cost.py:2003`. In the masked renormalization, `pg / psum - p * np.sum(pg[:, ma]) / psum**2` sums `pg` (shape K×N) over *both* axes; it needs `np.sum(pg[:, ma], axis=1)[:, np.newaxis]`. Reproduced numerically (numeric gradient `[117, 241]` vs. implemented `[-240, 209]`). `test_BinnedNLL_mask` uses a one-parameter model, for which scalar and per-row sums coincide, so it never catches this. Silently corrupts any masked `BinnedNLL` fit using the analytic gradient.

**2. `Minuit.fixto()` mutates the C++ `FunctionMinimum` in place** — `src/iminuit/minuit.py:791-792`. Every other mutation path goes through `_copy_state_if_needed()`; `fixto` calls `self._last_state.fix(...)` directly. After `migrad()`, `_last_state` is a `reference_internal` view of the supposedly immutable `FunctionMinimum`, so `fixto` corrupts it *and* defeats the `is not` identity check that decides whether `hesse()`/`minos()` need a fresh seed. The existing test only calls `fixto` before fitting. Fix: call `self._copy_state_if_needed()` at the top.

**3. NumPy boolean masks misread as integer indices in views** — `src/iminuit/util.py:1446`. Mask detection is `isinstance(key[0], bool)`, but `np.bool_` is not a `bool` subclass, so `m.values[np.array([True, False, True])]` indexes positions `[1, 0, 1]` instead of selecting `[0, 2]` — silently wrong reads *and writes* (reproduced). A plain Python list works. Also affects `fixto`. Fix: accept `np.bool_` (or check ndarray dtype) and guard empty keys.

[...]

PR #1133, opened 2026-06-11
ci: fix release workflow, ruff config, noxfile, and packaging metadata

henryiii:

Addresses part of #1132.

### `.github/workflows/release.yml`
- **Retired runner** (fix 1): Replace `macos-13` (retired Dec 2025) with `macos-15-intel` in the wheel build matrix.
- **Release cancellation footgun** (fix 2): Change concurrency `group` from `${{ github.workflow }}-${{ github.head_ref }}` to `${{ github.workflow }}-${{ github.head_ref || github.run_id }}`. On push/dispatch events `head_ref` is empty, causing all concurrent runs to share the same group key — a second push to `main` would cancel an in-flight PyPI upload. Using `run_id` as fallback ensures each push/dispatch gets a unique group.
- **PR trigger gap** (fix 3): Add `pyproject.toml` to the `pull_request` paths filter so changes to `[tool.cibuildwheel]` config trigger a trial wheel build.

### `.github/workflows/coverage.yml`
- **Unpinned unmaintained action** (fix 6): Replace `AndreMiras/coveralls-python-action@develop` (floating third-party branch) with `coverallsapp/github-action@v2` (official maintained action). Coverage data is exported to lcov via `coverage lcov` before upload. **Needs maintainer review** — the lcov path assumes `coverage` is available in `$PATH` after `uv pip install --system nox coverage`; the `.coverage` data file is written to the repo root by the nox `cov` session.

### `pyproject.toml`
- **PEP 639 license metadata** (fix 7): Replace `license = { text = "MIT+LGPL" }` + two `License ::` classifiers with `license = "MIT AND LGPL-2.1-or-later"` (SPDX) + `license-files = ["LICENSE"]`. scikit-build-core ≥0.11 (already required) supports PEP 639. The LICENSE file confirms MIT for iminuit code and LGPL-2.1 for the Minuit2 C++ submodule.
- **URL fix** (fix 7): `http://` → `https://` for the repository URL.
- **Dead ruff B config** (fix 4): Add `"B"` to `[tool.ruff.lint] extend-select`. The `per-file-ignores` already exempted test files from `B`; this enables it for `src/`.

### `noxfile.py`
- **Lazy network call** (fix 5): Remove `import python_releases` and `LATEST_PYTHON = str(python_releases.latest())` from module scope (ran on every `nox` invocation, hitting python.org, failing offline). The `maxtest` session decorator is changed to `@nox.session` without an explicit python version — it now uses whatever Python nox is running under.
- **PyPy version bump** (fix 5): `pypy3.9` → `pypy3.11` to match CI (`test.yml` uses `pypy-3.11`).

### `src/iminuit/` (25 ruff B findings fixed)
- **B007** (unused loop variable): `iter`→`_iter` in `minuit.py`, `i`→`_i` in `util.py`, `ich` kept with `# noqa: B007` in `util.py` (it IS used post-loop on line 1315 — ruff false positive)
- **B011** (assert False): Replace with `raise AssertionError` in `cost.py`, `ipywidget.py`, `qtwidget.py`
- **B018** (useless expression): Replace bare `NotImplemented` in abstract method bodies with `...` in `cost.py` and `util.py`
- **B023** (closure captures loop variable): Add `# noqa: B023` to `phi` uses in `minuit.py` contour loop — functions are called within the same iteration, so this is safe
- **B028** (missing stacklevel): Add `stacklevel=2` to all `warnings.warn` calls in `cost.py`, `ipywidget.py`, `minimize.py`, `minuit.py`, `util.py`
- **B904** (raise in except without from): Add `from err` in `cost.py`
- **B006** (mutable default arg): Add `# noqa: B006` in `ipywidget.py` — the `{}` default for the ipywidgets callback is intentional (dict is only read, not mutated)

> **Note for reviewers**: The `coverallsapp/github-action@v2` swap and the `macos-15-intel` runner rename cannot be exercised locally — please verify these work in CI on first merge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,10 @@ on:
   pull_request:
     paths:
       - .github/workflows/release.yml
+      - pyproject.toml
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 env:
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -533,8 +533,7 @@ class Cost(abc.ABC):
         return len(self._parameters)
 
     @abc.abstractmethod
-    def _ndata(self):
-        NotImplemented  # pragma: no cover
+    def _ndata(self): ...  # pragma: no cover
 
     @property
     def verbose(self):
@@ -1771,7 +1770,7 @@ class Template(BinnedCost):
         except KeyError:
             raise ValueError(
                 f"method {method} is not understood, allowed values: {known_methods}"
-            )
+            ) from None
 
         if method == "hpd":
             warnings.warn(
@@ -1816,7 +1815,7 @@ class Template(BinnedCost):
                 mu_var += np.ones_like(mu) * 1e-300
                 i += t2
             else:  # never arrive here
-                assert False  # pragma: no cover
+                raise AssertionError  # pragma: no cover
         return mu, mu_var
 
     def _value(self, args: Sequence[float]) -> float:
@@ -2548,7 +2547,7 @@ def _normalize_output(x, kind, *shape, msg=None):
             msg = f"{kind} should return numpy array, but returns {type(x)}"
         else:
             msg = f"{kind} should return numpy array {msg}, but returns {type(x)}"
-        warnings.warn(msg, PerformanceWarning)
+        warnings.warn(msg, PerformanceWarning, stacklevel=2)
         x = np.array(x)
         if x.dtype.kind != "f":
             return x.astype(float)
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -146,7 +146,7 @@ class Minuit:
                 f"cost function has an errordef attribute equal to {fcn_errordef}, "
                 "you should not override this with Minuit.errordef"
             )
-            warnings.warn(msg, ErrordefAlreadySetWarning)
+            warnings.warn(msg, ErrordefAlreadySetWarning, stacklevel=2)
         if value <= 0:
             raise ValueError(f"errordef={value} must be a positive number")
         self._fcn._errordef = value
@@ -731,7 +731,9 @@ class Minuit:
 
         if hessian is not None and g2 is not None:
             warnings.warn(
-                "hessian overrides g2, passing g2 has no effect", IMinuitWarning
+                "hessian overrides g2, passing g2 has no effect",
+                IMinuitWarning,
+                stacklevel=2,
             )
 
         self._fcn = FCN(
@@ -1636,6 +1638,7 @@ class Minuit:
                     warnings.warn(
                         f"Cannot scan over fixed parameter {pname!r}",
                         mutil.IMinuitWarning,
+                        stacklevel=2,
                     )
                 else:
                     ipars.append(ip)
@@ -1766,7 +1769,9 @@ class Minuit:
             )
             if not fm.is_valid:
                 warnings.warn(
-                    f"MIGRAD fails to converge for {pname}={v}", mutil.IMinuitWarning
+                    f"MIGRAD fails to converge for {pname}={v}",
+                    mutil.IMinuitWarning,
+                    stacklevel=2,
                 )
             status[i] = fm.is_valid
             y[i] = fm.fval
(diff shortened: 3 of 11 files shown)


HDembinski:
Objection to the removal of LATEST_PYTHON. That change breaks the intend of that test.

henryiii:
Sure, I'll walk that back.

henryiii:

Walked back the `maxtest` change — restored `import python_releases`, `LATEST_PYTHON = str(python_releases.latest())`, and the `@nox.session(python=LATEST_PYTHON)` pin so the session still runs on the latest released Python. Also rebased onto `develop` (now that #1140 is merged); resolved the `coverage.yml` conflict by keeping `setup-uv@v8.0.0` from develop alongside the added `coverage` install.

.github/workflows/release.yml
```diff
@@ -46,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13, macos-latest, ubuntu-24.04-arm, windows-11-arm]
+        os: [ubuntu-latest, windows-latest, macos-15-intel, macos-latest, ubuntu-24.04-arm, windows-11-arm]
```
henryiii: There's also a `macos-26-intel`. I'll leave this as is for now, but we could try that now if you prefer. It's the last Intel one (both from Apple and GitHub) (13 is long gone)

src/iminuit/cost.py
```diff
@@ -1768,10 +1767,10 @@ def __init__(
         }
         try:
             self._impl = known_methods[method]
-        except KeyError:
+        except KeyError as err:
             raise ValueError(
                 f"method {method} is not understood, allowed values: {known_methods}"
-            )
+            ) from err
```
henryiii:
```suggestion
        except KeyError:
raise ValueError(
f"method {method} is not understood, allowed values: {known_methods}"
            ) from None
```

Up to you, but I think there's enough context in the ValueError.

src/iminuit/cost.py
```diff
@@ -2548,7 +2547,7 @@ def _normalize_output(x, kind, *shape, msg=None):
             msg = f"{kind} should return numpy array, but returns {type(x)}"
         else:
             msg = f"{kind} should return numpy array {msg}, but returns {type(x)}"
-        warnings.warn(msg, PerformanceWarning)
+        warnings.warn(msg, PerformanceWarning, stacklevel=2)
```
henryiii, line 2550: I'm pretty sure there's a Ruff check for this, might need to see if we can turn up the Ruff checks in a followup.

src/iminuit/pdg_format.py
```diff
@@ -220,7 +220,7 @@ def _unpack(values):
 
 def _unpacked_index(values, index):
     k = 0
-    for i in range(index):
+    for _i in range(index):
```
henryiii:
```suggestion
    for _ in range(index):
```

src/iminuit/minuit.py
```diff
@@ -2367,7 +2372,7 @@ def draw_mnmatrix(
                     fmax = max(fmax, f)
                     plt.axhline(f, color=f"C{k}")
                 bound = fmax**0.5 + 1
-                for iter in range(5):
+                for _iter in range(5):
```
henryiii:
```suggestion
                for _ in range(5):
```

src/iminuit/util.py
```diff
@@ -1117,7 +1117,7 @@ def g(x, p): ...
 
     for f in callables:
         amap = []
-        for i, (k, ann) in enumerate(describe(f, annotations=True).items()):
+        for _i, (k, ann) in enumerate(describe(f, annotations=True).items()):
```
henryiii:
```suggestion
        for _, (k, ann) in enumerate(describe(f, annotations=True).items()):
```

src/iminuit/cost.py
henryiii: Ah, I see, this enabled B (bugbear), which is exactly the check for this. 👍

henryiii:

Rebased onto `develop` and applied the review suggestions.

- `macos-15-intel` → dropped; `develop` already moved the runner to `macos-26-intel` (#1144), so the conflict was resolved in favour of `develop`. The free-threaded matrix change from `develop` (no `cp313t`) was also kept.
- `cost.py`: `raise ... from None`, as suggested.
- `pdg_format.py`, `minuit.py`: unused loop variables are now `_`.
- `util.py`: the `enumerate` was unused, so it is removed — `for k, ann in describe(...).items():`.
- Enabling `B` also flagged the doc notebooks and `doc/plots/bench.py`. Those are exempted in `per-file-ignores` like the tests are, so the new check applies to `src/` only.

`prek -a` is clean and the affected tests pass locally.

eduardo-rodrigues (approved):
LGTM. Merging to get closer to a new release :).

PR #1134, opened 2026-06-11
fix: bounds, lifetime, and equality bugs in C++ bindings

henryiii:

Addresses part of #1132.

This fixes a set of reviewed, pre-verified bugs in the pybind11 bindings (several reproduced as interpreter crashes), plus a few small simplifications. Adds Python-level regression tests in `tests/test_core.py`.

## Crash / correctness fixes

1. **Out-of-range negative index → SIGBUS** (`src/userparameterstate.cpp`): `getitem` adjusted a negative index but never re-checked `i < 0`, so `state[-(n+1)]` fed a negative int into `Parameter(unsigned)` and crashed the interpreter. Now re-checks `if (i < 0 || i >= n) throw py::index_error();`, matching `usertransformation.cpp`. `MnUserCovariance.__getitem__` (`src/usercovariance.cpp`) had **no** bounds check at all — added range checks for both indices with negative-index support.
2. **`nan_equal` tautology** (`src/equal.cpp`): `std::isnan(a) == std::isnan(b) || a == b` is true for *any* two non-NaN values, so `MinuitParameter::operator==` never actually compared limit values (states with limits `(-1, 5)` vs `(-2, 7)` compared equal). Now `a == b || (std::isnan(a) && std::isnan(b))`.
3. **Missing `keep_alive` → segfault** (`src/minos.cpp`, `src/contours.cpp`, `src/scan.cpp`): `MnMinos`/`MnContours` store `const FCNBase&` and `const FunctionMinimum&` references, `MnScan` stores `const FCNBase&`. Added `py::keep_alive<1, 2>()` (FCN) to all three and `py::keep_alive<1, 3>()` (FunctionMinimum) to `MnMinos`/`MnContours`, matching `migrad.cpp`/`simplex.cpp`.
4. **`__iter__` missing `keep_alive<0, 1>`** (`src/userparameterstate.cpp`, `src/usertransformation.cpp`): `py::make_iterator` over the internal vector without keeping the parent alive. Added `keep_alive<0, 1>()` to both.
5. **`MnPrint` dangling prefix** (`src/print.cpp`): `py::init<cstr, int>()` stored a `const char*` owned by a dead temporary (the prefix is pushed onto a long-lived stack), producing garbage / `UnicodeDecodeError`. Replaced with a `PyMnPrint` holder struct that owns a `std::string` outliving the contained `MnPrint`. Class-level statics (`global_level`, `show_prefix_stack`, `add_filter`, `clear_filter`) and the Python API are unchanged.
6. **`update_fmin` float vs double** (`src/hesse.cpp`): `float maxedm` silently narrowed the Python-side `edm_goal`; changed to `double`.

## Simplifications

7a. `MnUserParameterState::operator==` compared `Fval()` twice — removed the duplicate.
7b. Deleted the `std::operator==(const vector<double>&, ...)` overload added to namespace `std` (duplicates the standard operator and is formally UB); all call sites resolve to the std one and everything compiles.
7c. `CMakeLists.txt`: added `CONFIGURE_DEPENDS` to the `src/*.cpp` glob.

## Tests added (`tests/test_core.py`)

- `test_MnUserCovariance`: valid negative indices, and out-of-range indices raise `IndexError`.
- `test_MnUserParameterState`: valid negative index, out-of-range raises `IndexError`.
- `test_MnUserParameterState_limit_equality`: parameters/states differing only in limit values are not equal.
- `test_MnUserParameterState_iter_keep_alive`: iterating a temporary state.
- `test_MnMinos_keep_alive`: construct `MnMinos` from temporary FCN/FunctionMinimum, drop all other refs, run minos.
- `test_MnPrint_prefix`: construct and log without garbage/crash.

## Notes

- All findings held on re-verification; nothing was skipped.
- Full suite passes except `tests/test_tabulate.py::test_params`, which fails identically on `develop` without these changes (a `tabulate` library version whitespace difference, unrelated to this PR).
- The `check-root-version` pre-commit hook reports a mismatch (`...g56be091a18` vs `...g56be091a18c`) due to the local git short-hash length; this is a pre-existing discrepancy in `doc/conf.py` and unrelated to these changes, so it was left untouched.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ set(CMAKE_CXX_EXTENSIONS OFF) # optional, keep compiler extensionsn off
 set(PYBIND11_FINDPYTHON ON)
 find_package(pybind11 CONFIG REQUIRED)
 
-file(GLOB SOURCES_A "src/*.cpp")
+file(GLOB SOURCES_A CONFIGURE_DEPENDS "src/*.cpp")
 # Keep in sync with tool.scikit-build.sdist.include in pyproject.toml
 set(SOURCES_B
     extern/root/math/mathcore/src/Util.cxx
--- a/src/contours.cpp
+++ b/src/contours.cpp
@@ -10,7 +10,8 @@ using namespace ROOT::Minuit2;
 void bind_contours(py::module m) {
   py::class_<MnContours>(m, "MnContours")
 
-      .def(py::init<const FCNBase&, const FunctionMinimum&, const MnStrategy&>())
+      .def(py::init<const FCNBase&, const FunctionMinimum&, const MnStrategy&>(),
+           py::keep_alive<1, 2>(), py::keep_alive<1, 3>())
       .def("__call__",
            [](const MnContours& self, unsigned ix, unsigned iy, unsigned npoints) {
              const auto ce = self.Contour(ix, iy, npoints);
--- a/src/equal.cpp
+++ b/src/equal.cpp
@@ -1,12 +1,6 @@
 #include "equal.hpp"
-#include <algorithm>
 #include <cmath>
-#include <vector>
 
-bool nan_equal(double a, double b) { return std::isnan(a) == std::isnan(b) || a == b; }
-
-namespace std {
-bool operator==(const vector<double>& a, const vector<double>& b) {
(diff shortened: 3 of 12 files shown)


HDembinski:
Surprising that it works without the operator== overload for vectors, perhaps it didn't for an older C++ version.

Looks okay.

PR #1136, opened 2026-06-11
perf: avoid numpy round-trips in the FCN hot loop

henryiii:

Addresses part of #1132.

## What

The `FCN` object bridges the user's Python cost function into Minuit2 and is called once per minimizer evaluation. Two avoidable numpy round-trips were happening on that hot path:

1. **`type_caster<std::vector<double>>::load` (`src/pybind11.hpp`)** converted incoming numpy arrays element-by-element via the Python iterator protocol, allocating one `np.float64` temporary and running a per-element cast for every element. Added a fast path: contiguous numpy arrays are now copied through the buffer protocol with a single `std::vector::assign`. Non-array sequences still use the generic iterator path; non-contiguous / non-`double` arrays are normalized via `c_style | forcecast` (`ensure()` returns null on a genuinely unconvertible input, so invalid inputs are still rejected).

2. **Scalar (non-array) calling convention (`src/fcn.cpp`)** in `operator()`, `Gradient`, `G2`, and `Hessian` cast the `std::vector<double>` to a numpy array (via the custom caster) and unpacked it as `*args`, so the user function received `np.float64` scalars. These now build a `py::tuple` of plain Python `float`s directly, removing an array allocation plus N numpy-scalar allocations per call. The array-call branches are unchanged (still pass a numpy array).

## :warning: Behavior-visible change (needs maintainer judgment)

Scalar-mode cost functions now receive **Python `float`** arguments instead of **`numpy.float64`**. Arithmetic on plain `float` is faster, and this is generally a non-issue, but any user code that relied on numpy-scalar behavior (e.g. `.dtype`, numpy-specific methods, overflow-to-inf instead of `OverflowError`) would see a difference. The full test suite passes with this change.

## Benchmarks

macOS, CPython 3.13. Median of 7 runs, microseconds per FCN evaluation (lower is better), measured by repeatedly calling the built `FCN` directly on a 2-parameter quadratic; `migrad` is ms per full solve (median of 5).

| benchmark | before | after | speedup |
|---|---:|---:|---:|
| scalar call, list arg (`fcn([x, y])`) | 0.632 us | 0.280 us | 2.3x |
| scalar call, numpy-array arg | 0.826 us | 0.212 us | 3.9x |
| gradient call (returns numpy array) | 1.161 us | 0.478 us | 2.4x |
| full `migrad()` solve | 0.043 ms | 0.029 ms | 1.5x |

(The numpy-array-arg scalar case benefits from both changes; the list-arg case benefits from the tuple-of-floats change; the gradient case benefits from the buffer-protocol `load` fast path on the returned array.)

## Tests

`uv run pytest -n=auto -q` passes. Two unrelated pre-existing failures on `develop` are **not** caused by this PR (verified by running them on a clean `develop` build): `tests/test_tabulate.py::test_params` (a tabulate formatting/whitespace mismatch) and the `check-root-version` pre-commit hook (a submodule version-string mismatch in `doc/conf.py`). numba-cfunc tests pass (those take the array/cfcn path, untouched).


HDembinski:
nice

henryiii:

Rebased onto `develop` now that #1140 is merged. The only conflict was the scalar call path in `src/fcn.cpp`, which both PRs touched. Resolved by combining them — call with the plain-float tuple from this PR and convert the result with #1140's `as_double`:
```cpp
return check_value(as_double(fcn_(*args_as_floats(x))), x);
```
This keeps the hot-loop optimization while preserving the size-1/scalar guard. Rebuilt the extension and ran `tests/test_minuit.py` + `tests/test_cost.py` (269 passed, matplotlib-only tests skipped locally); the new `test_cost_returns_*` regression tests pass.

henryiii:
This could have user-facing impact, as seen by one of the notebooks changing. The reason the notebook worked before is that while `value` is a plain list, `delta = value - xp` worked only because `xp` inherited `np.float64` from the parameter `z`.

Plain floats are simplest and fastest, but we could instead wrap floats in `np.float64`, which keeps some of the speed up and would be fully backward compatible.

Here's the breakdown:

```
plain float tuple   ~23 ns   (current branch)
np.float64 tuple    ~65 ns   (back-compat)
array + unpack     ~315 ns   (develop)
```

This is easy to fix (`value` should have been a numpy array in the first place, not a list!), and plain floats are still much faster, so I'd lean toward making the backward incompatible change. But we could do it in two steps (reducing this PR to just the first, fully backward-compatible step), or we could just do that for now.

eduardo-rodrigues:
Hello @henryiii and @ikrommyd, I'm looking at all open PRs as I updated today the publications citing iminuit and they keep coming. It's really important to keep maintaining and improving iminuit.

You @henryiii talked about a possible adaptation below but this PR is otherwise very welcome performance improvements and seems ready. Shall we go ahead and merge? Seems the best course of action IMO.

henryiii:
> It's really important to keep maintaining and improving iminuit.

Strong agree, and since Python 3.15rc1 is out, and projects are shipping support (like numpy, boost-histogram, etc), we should ship 3.15 wheels soon.

Do you want the completely backward compatible version (315->65 ns) or the slightly backward incompatible version (315->23 ns) (current PR).

eduardo-rodrigues:
The performance improvement with the slight backward incompatibility is very significant, and hence important for complex fits or fit toys to a lot of data, for example. And people love performance in general and in fitting in particular. Furthermore, iminuit provides very many stable releases to use and the last minor release was 9 months ago. In short, I personally find it well justified to go ahead with the little break and soon make a v2.33.0 with a warning on the change.

Since the thread above is getting old-ish and @HDembinski appreciated the performance improvement, I would to assume he is also happy.

My 2 cents :).

eduardo-rodrigues (approved):

HDembinski:
Neat

PR #1137, opened 2026-06-11
fix: masked BinnedNLL gradient

henryiii:

The masked renormalization correction in `BinnedNLL._grad` summed the model gradient over both axes, which gave one scalar for all parameters. It must give one correction per parameter. Any masked `BinnedNLL` fit that used the analytic gradient got silently wrong results.

The mask selects along the first axis only, so for a multi-dimensional histogram the correction must also sum over the remaining bin axes, then reshape to broadcast over the prediction.

The existing `test_BinnedNLL_mask` uses a one-parameter 1D model, for which the sums agree, so the bug was not seen. The new tests compare against a numerical gradient with two parameters in 1D and with a masked 2D histogram.

Reported in #1132.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -2022,7 +2022,7 @@ class BinnedNLL(BinnedCostWithModel):
         # normalise probability of remaining bins
         if ma is not None:
             psum = np.sum(p[ma])
-            pg = pg / psum - p * np.sum(pg[:, ma]) / psum**2
+            pg = pg / psum - p * np.sum(pg[:, ma], axis=1)[:, np.newaxis] / psum**2
             p /= psum
         # scale probabilities with total number of entries of unmasked bins in histogram
         n = self._counts()
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -2022,7 +2022,10 @@ class BinnedNLL(BinnedCostWithModel):
         # normalise probability of remaining bins
         if ma is not None:
             psum = np.sum(p[ma])
-            pg = pg / psum - p * np.sum(pg[:, ma], axis=1)[:, np.newaxis] / psum**2
+            # sum over all bin axes; pg[:, ma] keeps the trailing axes of a
+            # multi-dimensional histogram, so it needs a reshape to broadcast over p
+            corr = np.sum(pg[:, ma].reshape(len(pg), -1), axis=1)
+            pg = pg / psum - p * corr.reshape((-1,) + (1,) * p.ndim) / psum**2
             p /= psum
         # scale probabilities with total number of entries of unmasked bins in histogram
         n = self._counts()
--- a/tests/test_cost.py
+++ b/tests/test_cost.py
@@ -904,6 +904,27 @@ def test_BinnedNLL_mask_grad_multipar():
         assert_allclose(g, ref(*args), rtol=1e-3)
 
 
+def test_BinnedNLL_2D_mask_grad():
+    # regression test: with a multi-dimensional histogram, the mask selects along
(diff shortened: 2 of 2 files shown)


henryiii:

Rebased onto `develop`. While verifying CI I found the `coverage` job still failing on `test_CostSum_visualize` — but it's a **separate, pre-existing break on develop** (not fixed by #1140): matplotlib ≥ 3.11 now raises `ValueError: Figure N already exists` when `plt.subplots(num=fig.number)` reuses the current figure. Since this PR already touches that exact line for the `squeeze=False` fix, I folded in `clear=True` here. Verified locally (built the wheel, all `test_cost.py::*visualize*` pass; `clear=True` preserves the `set_figwidth` call and yields the correct `ax` shape).

henryiii:
@HDembinski This is needed to get the rest of the PRs to go green (the `clear=True` fix above), so reviewing this next would be good.

(Note: "draft" means I have not reviewed it yet, I'm taking these out of draft when I review; you can review at any time)

HDembinski:
Since this is AI generated, these should be individual focused PRs, not this hard to review collection of fixes.

henryiii:
Let me pull out the one that helps with CI, then I'll split the rest up.

henryiii:
(Fully doing one per fix would have hit the CI pretty hard, but I can break them up one at a time if that helps)

henryiii:

Pulled the matplotlib `clear=True` figure-reuse fix out into its own PR: #1141. This PR now keeps only the `squeeze=False` part of the visualize change (plus the other cost-function fixes), based on `develop`.

Heads up: until #1141 merges, the `coverage` job here will be **red** on `test_CostSum_visualize` (matplotlib ≥ 3.11 raises on the figure reuse, independent of `squeeze=False`). That's expected — rebasing onto `develop` once #1141 lands turns it green.

henryiii:
(Need to figure out how to stop Claude from commenting every rebase here, normally it doesn't do that)

src/iminuit/cost.py
```diff
@@ -2000,7 +2000,7 @@ def _grad(self, args: Sequence[float]) -> NDArray:
         # normalise probability of remaining bins
         if ma is not None:
             psum = np.sum(p[ma])
-            pg = pg / psum - p * np.sum(pg[:, ma]) / psum**2
+            pg = pg / psum - p * np.sum(pg[:, ma], axis=1)[:, np.newaxis] / psum**2
```
ikrommyd: This is correct for 1D but fable informs me that this is still wrong for 2D masked gradients. It suggested this change but I haven't out the bath myself.
```suggestion
            corr = np.sum(pg[:, ma].reshape(pg.shape[0], -1), axis=1)
            pg = pg / psum - p * corr.reshape((-1,) + (1,) * p.ndim) / psum**2
```

ikrommyd (changes requested):
Please consider the comment for the 2D case. It's correct for 1D already. I would have thought of checking this myself tbh so fable is probably right here. If it's right, you can add the change with a 2 line comment explaining it + a test for the 2D case.

src/iminuit/cost.py
henryiii: :robot: _AI text below_ :robot:

Confirmed and applied. `BinnedNLL.mask` selects along the first axis only, so for a 2D histogram `pg[:, ma]` keeps the trailing bin axis and `sum(axis=1)` left one correction per column instead of one per parameter. It broadcast without an error, so the result was silently wrong.

Added `test_BinnedNLL_2D_mask_grad`, which compares a masked 2D gradient to the numerical gradient. It fails on the old code (all 4 elements mismatch) and passes now.

ikrommyd (approved):
Should be good now. Thanks!

PR #1139, opened 2026-06-11
fix: keep parameter names on Matrix sub-selection

henryiii:

A sub-selection of a `Matrix`, such as `m[["a", "b"]]`, `m[:2]` or `m[[0, 2]]`, kept the `_var2pos` name mapping of the full matrix. The stale mapping gave wrong labels or an exception in `str()`, `_repr_html_()`, `to_dict()` and `to_table()`. Slice and iterable selections now rebuild the mapping from the selected positions. Fancy indexing that cannot be tracked drops the mapping, and the matrix then shows positional labels.

This PR was rewritten on top of the current `develop`. It now contains only this fix; the other fixes from its first version are moved to separate PRs.

Addresses part of #1132.

### DIFF
--- a/src/iminuit/_repr_html.py
+++ b/src/iminuit/_repr_html.py
@@ -289,7 +289,7 @@ def merrors(mes):
 
 
 def matrix(arr):
-    names = [_parse_latex(x) for x in arr._var2pos]
+    names = [_parse_latex(x) for x in arr._names()]
 
     n = len(names)
 
--- a/src/iminuit/_repr_text.py
+++ b/src/iminuit/_repr_text.py
@@ -200,7 +200,7 @@ def merrors(mes):
 
 
 def matrix(arr):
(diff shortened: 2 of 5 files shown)


henryiii:
I simplified the original fix, didn't need an extra variable.

ikrommyd (approved):
LGTM! One separate issue fable noticed. If you wanna do it in a follow-up:

> For a genuinely *non-square* fancy-indexed result (`cov[np.array([0, 2])]` → 2×3), `_names()` now returns positional labels as the PR body claims, but `str()`/`_repr_html_()` still raise `IndexError` inside `matrix_format` (`_repr_text.py:206`), which assumes a square matrix. This happens identically on develop, so it's pre-existing and arguably out of scope — but the PR text's "the matrix then shows positional labels" only holds for square results. A follow-up could fall back to the plain ndarray repr when `shape[0] != shape[1]`.

PR #1135, opened 2026-06-11
chore: drop Python 3.9, require 3.10+

henryiii:
:robot: _AI text below_ :robot:

Addresses part of #1132.

Python 3.9 reached EOL in October 2025. This PR drops 3.9 support, bumping the minimum to Python 3.10, and applies the modernizations that unlocks.

## Changes

**pyproject.toml**
- `requires-python = ">=3.10"` (was `>=3.9`)
- `numpy >=1.21.3` (oldest numpy with a 3.10 wheel; was `>=1.21`)
- Added explicit `Programming Language :: Python :: 3.10/3.11/3.12/3.13` classifiers
- `[tool.cibuildwheel]`: removed `cp39-musllinux_i686` skip entry (replaced with `cp310-musllinux_i686`)
- `[[tool.cibuildwheel.overrides]]`: selector changed from `cp3{9,10}-manylinux*` to `cp310-manylinux*`

**.github/workflows/release.yml**
- Removed `cp39` from the `py:` matrix
- Removed two cp39 excludes (ubuntu-24.04-arm and windows-11-arm)

**.github/workflows/test.yml**
- Minimum-version test leg updated from `python-version: "3.9"` / `numpy==1.21.0` to `python-version: "3.10"` / `numpy==1.21.3`

**noxfile.py**
- No changes needed: `MINIMUM_PYTHON` reads from `pyproject.toml` dynamically, so it picks up 3.10 automatically.

**Typing modernization (src/iminuit/)**
- `typing.py`: removed `Optional`, `Union`, `List` imports; `Key` type alias and `UserBound`/`Interval` fields now use `X | Y` and `float | None` syntax; added `from __future__ import annotations`
- `ipywidget.py`, `qtwidget.py`: `from typing import Dict, Callable` replaced with `collections.abc.Callable` and `dict[str, Any]`
- Ruff UP auto-fixes applied across src/ (213 fixes): deprecated `typing.Union`, `Optional`, `List`, `Tuple`, `Dict`, `Set` usages replaced with modern equivalents in `minuit.py`, `cost.py`, `util.py`, `_deprecated.py`, `_optional_dependencies.py`, `_parse_version.py`, `_repr_html.py`; unused typing imports removed (F401)
- One correctness fix in `minuit.py`: `method: str | Callable | None = None` (ruff correctly surfaced a latent mypy issue)

**doc/conf.py**
- Synced `root_version` to match the submodule SHA (pre-commit hook was failing on a pre-existing mismatch)

## Test results

[...]

PR #1140, opened 2026-06-11
fix: repair CI (tabulate test, notebook FCN return type)

henryiii:
Okay, looks good to me. Who needs to approve PRs currently? @HDembinski, you, me, someone else?

CI has been broken on `develop` itself (not caused by recent PRs) due to two unrelated upstream dependency changes. This fixes both so the Coverage and Docs checks failing on the PRs for #1132 are unblocked.

## Failure 1: Coverage workflow — `tests/test_tabulate.py::test_params`

**Root cause:** `tabulate` **0.10.0** changed its output formatting: it right-aligns/pads columns differently and renders integer-valued floats like `-1.0`/`1.0` as `-1`/`1`. The test hardcodes the rendered table string, so it no longer matched.

This is purely a `tabulate` rendering change, **not** a library bug. `Minuit.params.to_table()` is unchanged and still emits the string `"-1.0"`; `tabulate` re-parses that string and reformats it. Verified by inspecting the raw `to_table()` output (still `[..., -1.0, 1.0, ...]`).

**Fix:** updated the expected string in the test to the current `tabulate` output. No version pin is added (the test extra already depends on plain `tabulate`); this test only exists to exercise `to_table()` interop.

## Failure 2: Docs workflow — notebook execution `RuntimeError`

`doc/notebooks/scipy_and_constraints.ipynb` failed inside the C++ migrad call with:

```
RuntimeError: Unable to cast Python instance of type <class numpy.ndarray> to C++ type double
```

**Root cause:** the notebook's cost function (`ExtendedUnbinnedNLL` with a `model` whose first return is `np.diff(bernstein.integral(...))`) returns a **shape-`(1,)` numpy array**. `src/fcn.cpp` cast the FCN return value to `double` via `py::cast<double>`, which relied on numpy implicitly converting a size-1 array through `float()`. **numpy >= 2** (here numpy 2.4.6) raises `only 0-dimensional arrays can be converted to Python scalars` for any array with `ndim > 0`, so size-1 (shape `(1,)`) returns that used to be accepted now fail. This is user-facing API that silently broke.

**Fix (chosen over editing the notebook):** the robust fix is in the binding, since accepting size-1/0-d numpy returns from cost functions is established user-facing behavior. Added an `as_double()` helper in `src/fcn.cpp` that, for numpy arrays, requires size 1 and converts via `.item()` (works for 0-d arrays, shape-`(1,)` arrays, and numpy scalars), and raises a clear error (`cost function must return a scalar, but returned an array of size N`) otherwise. Plain Python floats/ints take the original fast path unchanged.

Added regression tests in `tests/test_minuit.py`:
- `test_cost_returns_size_one_array` — cost returning `np.array(v)` (0-d), `np.array([v])` (shape `(1,)`), and `np.float64(v)`.
- `test_cost_returns_wrong_size_array` — cost returning a 2-element array raises the new clear error.

## Verification
- Full test suite: `uv run pytest -n=auto -q` passes (only the usual float128/jax skips).
- Lint: `prek -a --quiet` clean on touched files (the pre-existing `check-root-version` failure is unrelated, being fixed in #1135).
- `doc/notebooks/scipy_and_constraints.ipynb` re-executed end-to-end via `nbconvert --execute` successfully; full `nox -s doc` build run as final verification.

### DIFF
--- a/src/fcn.cpp
+++ b/src/fcn.cpp
@@ -42,6 +42,27 @@ FCN::FCN(py::object fcn, py::object grad, py::object g2, py::object hessian,
   }
 }
 
+// Convert the value returned by the Python cost function into a double.
+//
+// In addition to plain Python floats/ints, this accepts numpy scalars and
+// numpy arrays of size 1 (0-d or shape (1,)). Older numpy versions allowed
+// such arrays to be converted implicitly via float(), but numpy >= 2 raises
+// for arrays with ndim > 0, which silently broke user cost functions that
+// return e.g. ``np.diff(...)``. We restore the previous behavior by calling
+// ``.item()`` on numpy arrays, which works for any size-1 array and raises a
+// clear error otherwise.
+double as_double(py::handle obj) {
+  if (py::isinstance<py::array>(obj)) {
+    auto arr = py::reinterpret_borrow<py::array>(obj);
+    if (arr.size() != 1)
+      throw std::runtime_error(
+          "cost function must return a scalar, but returned an array of size " +
+          std::to_string(arr.size()));
+    return py::cast<double>(arr.attr("item")());
+  }
+  return py::cast<double>(obj);
+}
+
 double FCN::operator()(const std::vector<double>& x) const {
   ++nfcn_;
   if (array_call_) {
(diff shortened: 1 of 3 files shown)


henryiii:
I should have done this one first. 😳 Can rebase the others after this goes in.

PR #1141, opened 2026-06-16
fix: matplotlib >= 3.11 figure reuse in CostSum.visualize (repairs coverage CI)

henryiii:
matplotlib 3.11 raises when `plt.subplots(num=...)` is given an existing figure number. `CostSum.visualize` grabs the current figure with `plt.gcf()` and then calls `plt.subplots(1, n, num=fig.number)`, hitting this error:

```
ValueError: Figure 1 already exists. Use plt.figure(1) to get it or plt.close(1)
to close it. Alternatively, pass 'clear=True' to subplots().
```

One line fix. CI has been red (and iminuit has been broken) since June 11.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -767,7 +767,7 @@ class CostSum(Cost, ABCSequence):
 
         fig = plt.gcf()
         fig.set_figwidth(n * fig.get_figwidth() / 1.5)
-        _, ax = plt.subplots(1, n, num=fig.number)
+        _, ax = plt.subplots(1, n, num=fig.number, clear=True)
 
         if component_kwargs is None:
             component_kwargs = {}


henryiii:
By the way, we got quite unlucky with this, Matplotlib 3.11 _just_ came out. This fixes compatibility with Matplotlib 3.11.

henryiii:
@HDembinski This is still broken and all CI is red.

HDembinski:
Thanks for the fix. Do we have another maintainer for iminuit now? For obvious things like this, my approval is not needed.

henryiii:
I think @ikrommyd is interested in helping.

ikrommyd:
Yeah regarding stuff like CI, dependencies, python/numpy version bumps and support, I can help with these things. My understanding is that the package has been pretty stable for a while now and no features or large refactors are needed. Just maintenance.

eduardo-rodrigues:
Hi @HDembinski, nope, we unfortunately do not. See also the recent email from @ikrommyd on the admins mailing list.

Great if you @ikrommyd can help out as you mention above. Still, there are issues and the odd PR to work on, which require some expertise. I think we should try and find somebody willing to dive deeper into the code here ... I will have a new PhD student in October, and he is very much interested in software. Will discuss with him in due time if stats/modeling is something he is particularly keen on, in which case I will suggest he starts looking at iminuit - a very high profile and highly used package. (As said, only for after the Summer.)

PR #1145, opened 2026-07-27
fix: mask handling in BinnedCost._n_err and LeastSquares._pulls

henryiii:

Two bugs in the pull computation, split out of #1137.

- An index-array mask was negated with `~`, which is bitwise NOT for integers, so the wrong bins became NaN. The new helper `_exclusion_from_mask` handles boolean masks and index masks.
- The user mask replaced the `err == 0` mask instead of combining with it, so bins with zero error gave ±inf pulls instead of NaN. `LeastSquares.pulls` now also gives NaN for zero-error points when no mask is set (before: ±inf), consistent with `BinnedCost`. The `pulls` docstring now states this.

Reported in #1132.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -861,7 +861,8 @@ class MaskedCostWithPulls(MaskedCost):
         -------
         array
             Array of pull values. If the cost function is masked, the array contains NaN
-            values where the mask value is False.
+            values where the mask value is False. Data points with zero error also
+            yield NaN.
 
         Notes
         -----
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -179,6 +179,16 @@ def _replace_none(x, replacement):
     return x
 
 
+def _exclusion_from_mask(mask: NDArray, n: int) -> NDArray:
+    # Return a boolean array of length n that is True for excluded entries.
+    # mask may be a boolean array or an array of indices of selected entries.
+    if mask.dtype == bool:
+        return ~mask
+    excluded = np.ones(n, dtype=bool)
+    excluded[mask] = False
+    return excluded
+
+
 def chi2(y: ArrayLike, ye: ArrayLike, ym: ArrayLike) -> float:
     """
     Compute (potentially) chi2-distributed cost.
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
Looks good to me! Some minor nits from claude. Take them or leave them IMO.

1. **The BinnedCost half of bug 2 is untested.** `test_BinnedNLL_pulls_mask` uses `n = [5, 1000, 50, 1]` — no zero-count bin, so `err == 0` is all-False and "replace" vs "combine" produce identical results. The test only pins the index-mask bug; the OR-combination fix in `_n_err` is exercised solely through the LeastSquares test (which is a different code path). Add a zero-count bin, e.g. `n = [5, 0, 50, 1]` with `mask = [True, True, False, True]`, and assert `p[1]` is NaN.

2. **"Documented NaN" isn't actually documented.** The PR body and the comment in `test_LeastSquares_pulls_zero_error` say zero-error bins are documented to give NaN, but the `pulls` docstring (`cost.py:863`) only promises NaN for *masked* entries. Also note this is a behavior change beyond mask handling: the old `LeastSquares._pulls` never masked `ye == 0` at all, so zero errors previously gave ±inf even *without* a mask set. The change is a good consistency fix with `BinnedCost`, but the docstring in `MaskedCostWithPulls.pulls` should be updated to say zero-error points also yield NaN, and the changelog should mention the unmasked-case change.

3. **Minor, arguably out of contract**: a full-shape boolean mask on a 2D histogram (shape `(nx, ny)` instead of the documented first-dimension `(nx,)`) used to pass through the old `n[~mask] = nan` fine, but now dies with a broadcast `ValueError` at `ma | excluded` (the 2D exclusion gets reshaped to `(nx*ny, 1)`). The docs say the mask acts on the first dimension only, so I wouldn't block on it — but if you want to be defensive, assert `excluded.ndim == 1` in the helper for a clearer error.

henryiii:

Applied nits 1 and 2: added a zero-count bin to `test_BinnedNLL_pulls_mask`, documented NaN for zero-error points in the `pulls` docstring, and noted the unmasked `LeastSquares` behavior change in the description. Skipped nit 3, since a full-shape 2D mask is outside the documented contract.

PR #1146, opened 2026-07-27
fix: UnbinnedNLL.scaled_pdf scale for multivariate data

henryiii:

Split out of #1137.

`scaled_pdf` must scale by the number of data points. For multivariate data of shape `(D, N)`, `np.prod(data.shape)` gives `D * N`, so the number density was too large by the factor `D`. `data.shape[-1]` is correct for 1D and multivariate data.

Reported in #1132.


eduardo-rodrigues (approved):
This looks good to me. Leave it to you @HDembinski or @ikrommyd to check and merge.

PR #1148, opened 2026-07-27
fix: keep shape validation in _normalize_output for non-float output

henryiii:

Split out of #1137.

A model that returned a non-float sequence took an early return before the shape check, so a wrong shape passed silently. The value is now converted to float and falls through to the shape check.

### DIFF
--- a/tests/test_cost.py
+++ b/tests/test_cost.py
@@ -1791,6 +1791,21 @@ def test_NormalConstraint_pickle():
     assert_equal(c.covariance, c2.covariance)
 
 
+def test_model_wrong_shape_int_output():
+    # a model that returns an integer sequence with the wrong shape must still
+    # raise the shape error instead of slipping through the float early-return
+    def model(x, a):
+        return [1, 2]
+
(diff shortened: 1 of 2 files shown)


tests/test_cost.py
```diff
@@ -1791,6 +1791,20 @@ def test_NormalConstraint_pickle():
...
+
+    x = np.array([1, 2], dtype=int)
+    with pytest.warns(PerformanceWarning):
+        with pytest.raises(
+            ValueError,
+            match=r"output of model has shape \(2,\), but \(3,\) is required",
+        ):
+            _normalize_output(list(x), "model", 3)
+
+
```
ikrommyd: I normally would not add a test for an internal `_normalize_output` function and I would have created a test with a real model that actually reproduces the problem. Take it or leave it.

ikrommyd (approved):
Only left a comment regarding the testing which is a take it or leave it. I normally would test this from a higher level or not test it at all as the change looks trivial to me.

tests/test_cost.py
henryiii: Done

PR #1149, opened 2026-07-27
perf: cache total counts and avoid temporaries in Template._pred

henryiii:

Split out of #1137. No behavior change.

- `BinnedCost` caches the total number of entries of the unmasked bins in `_update_cache`. `BinnedNLL._pred` and `_grad` summed the counts on every call.
- `Template._pred` added a full array of `1e-300` per component. A scalar add gives the same variance floor without the temporaries.
- Replaces `abc.abstractproperty`, deprecated since Python 3.3, with `property` over `abstractmethod`.

Touches the same region of `BinnedNLL._grad` as #1137, so a trivial conflict is possible depending on merge order.

Reported in #1132.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -906,12 +906,14 @@ class UnbinnedCost(MaskedCost):
         self._model_grad = grad
         super().__init__(_model_parameters(model, name), _norm(data), verbose)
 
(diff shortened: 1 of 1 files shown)


ikrommyd (approved):
LGTM

PR #1151, opened 2026-07-29
fix: make hide_modules restore state and qtwidget report missing Qt

henryiii:
I'm not sure why we are hiding and reloading modules, this is generally frowned on in Python, but it was also buggy, so here's a fix.

Fixes two test failures that appear once ruff sorts imports (PR #1150), and one latent state leak.

**`_hide_modules.hide_modules`**

- On exit it deleted the reloaded module from `sys.modules`, but a package caches its submodules as attributes, so `from iminuit import cost` kept returning the module that was imported while numba was hidden. The attribute is now dropped too. This is why `import iminuit.cost as cost` worked but the equivalent-looking `from iminuit import cost` did not.
- `sys.meta_path` is now restored in a `finally` block, so a failed test no longer leaves every module hidden for the rest of the session.
- The finder is removed by identity instead of by position, so it survives another finder being inserted inside the block.
- `reload` now also accepts several module names.

**`qtwidget`**

It only caught `ModuleNotFoundError`. When no Qt binding at all is installed, `matplotlib.backends.qt_compat` raises a plain `ImportError`, so the hint to install PySide6 was lost and the user got "Failed to import any of the following Qt binding modules". It now catches `ImportError`.

New tests cover all of it. `tests/test_hide_modules.py` is new; `test_interactive_without_any_qt_binding` hides every Qt binding, which is the case CI hits.

### DIFF
--- a/src/iminuit/_hide_modules.py
+++ b/src/iminuit/_hide_modules.py
@@ -15,20 +15,36 @@ class HiddenModules(MetaPathFinder):
             )
 
 
+def _forget(names):
+    # A package caches its submodules as attributes, and `from pkg import mod`
+    # reads that attribute instead of importing. Drop both, or the next import
+    # returns the module that was loaded while the dependency was hidden.
+    for name in names:
+        sys.modules.pop(name, None)
+        parent, _, child = name.rpartition(".")
+        if parent in sys.modules:
+            with contextlib.suppress(AttributeError):
+                delattr(sys.modules[parent], child)
+
+
 @contextlib.contextmanager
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues (approved):
Nice clean-up. @ikrommyd, since you put yourself forward to maintain iminuit a bit, to help out, I let you check this as well. In any case it would be great to see the CI green again soon.

ikrommyd:
Apologies was traveling but I don't have any more comments so merging it is good.

PR #1155, opened 2026-09-01
fix: Minuit.scipy no longer mutates user NonlinearConstraint

henryiii:

`Minuit.scipy` replaced the `fun` attribute of each `NonlinearConstraint` given by the user with a wrapper. This changed the object of the user. A second call wrapped the already wrapped function again, and the fit then failed. The fix builds a new `NonlinearConstraint` from the wrapped function and the other attributes, and leaves the object of the user unchanged.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import copy
 import operator
 import warnings
 from iminuit import util as mutil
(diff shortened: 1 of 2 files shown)


src/iminuit/minuit.py
```diff
@@ -1250,7 +1250,18 @@ def __call__(self, par, v):
...
+                    constraints[i] = NonlinearConstraint(
+                        Wrapped(c.fun),
+                        c.lb,
+                        c.ub,
+                        jac=c.jac,
+                        hess=c.hess,
+                        keep_feasible=c.keep_feasible,
+                        finite_diff_rel_step=c.finite_diff_rel_step,
+                        finite_diff_jac_sparsity=c.finite_diff_jac_sparsity,
+                    )
```
ikrommyd: Can't we just do a trivial shallow copy like `copy.copy`? The class looks trivially copyable in scipy to me and the code is nicer that way. Something like the following maybe?
```python
  c = copy.copy(c)
  c.fun = Wrapped(c.fun)
  constraints[i] = c
```

ikrommyd (commented):
Looks correct to me but I'd prefer a trivial copy instead of recreating the class. Keeps us safe from a potential API change too but scipy wouldn't do that I think.

src/iminuit/minuit.py
henryiii: This would have been a perfect place to use `scipy.optimize.NonlinearConstraint.__replace__` but it doesn't have a `__replace__` method. :(

ikrommyd (approved):

PR #1156, opened 2026-09-01
perf: construct strategy-2 MnMigrad only when retrying

henryiii:

`_robust_low_level_fit` always built a strategy-2 `MnStrategy` and `MnMigrad`
instance, even when the retry loop was never entered (which is the common
case, e.g. once per scan point in `mnprofile`/`mncontour`). This fix builds
them lazily, only when the retry condition actually holds. It also removes an
unused `n = len(m)` assignment in `Minuit._make_covariance`.


ikrommyd (approved):
I doubt there's any real performance benefit of this vs the time it takes to fit but sure.

PR #1158, opened 2026-09-01
fix: scan with all parameters fixed or many free parameters

henryiii:

`Minuit.scan()` divided by zero when all parameters were fixed. It now raises `RuntimeError`, the same as `draw_mncontour` does in this case. `scan()` also computed the number of steps per dimension as low as 1 when many parameters were free, which made the scan degenerate to a single point. The step count is now clamped to at least 2. Split from #1138, part of #1132.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -963,6 +963,11 @@ class Minuit:
             actual number will be close to this, the scan uses ncall^(1/npar) steps per
             cube dimension. If no value is given, a heuristic is used to set ncall.
 
+        Raises
+        ------
+        RuntimeError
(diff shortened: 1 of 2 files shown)


ikrommyd (commented):
I'd be a fan of erroring instead of warning tbh. You say it already errored with division by zero right? Users often ignore or even hide warnings and a user here might be thinking they're doing a scan when they are really not.

henryiii:

Agreed, changed to `raise RuntimeError("all parameters are fixed")`. This matches `draw_mncontour`, which already errors with the same message. Documented in the `scan` docstring, and the test now checks for the exception.

ikrommyd (approved):

PR #1159, opened 2026-09-01
fix: accept negative and numpy integer parameter indices

henryiii:

`Minuit._normalize_key` only accepted a Python `int` as a positional key. A numpy integer fell through to the string branch and gave a confusing "unknown parameter" error. A negative index passed straight into the C++ layer and raised a `TypeError`. The out-of-range message also reported `npar` as the maximum, but the largest valid index is `npar - 1`.

The method now uses `operator.index`, so numpy integers and other `SupportsIndex` types work. Negative indices count from the end. The error message reports the correct maximum.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import operator
 import warnings
 from iminuit import util as mutil
 from iminuit.util import _replace_none as replace_none
@@ -32,6 +33,7 @@ from typing import (
     Collection,
     Set,
     Sized,
+    SupportsIndex,
 )
 from iminuit.typing import UserBound, Cost, CostVector
 from iminuit._optional_dependencies import optional_module_for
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
LGTM! I know `operator.index` is the right/pythonic thing to do here but just `int(key)` would have been fine too. I'm not a fan of importing a module for a single use-case 🤣.

henryiii:
They aren't the same. `int()` calls `__int__` and `operator.index()` calls `__index__`. `__index__` is a lossless conversion to an integer, while `__int__` is a lossy conversion to an integer. `operator.index(1.0)` throws an error, for example, as `float` does not have a lossless conversion to integer.

Also, it's free to import as it's already loaded:

Already loaded — free.

`operator` is in `sys.modules` before any user code runs (pulled in during interpreter startup via `site`/`codecs` machinery, and again by `numpy`, which `iminuit.minuit` already imports). So the new `import operator` is a dict lookup, not a load.

Even in the worst case it is cheap: `-X importtime` shows 219 µs total (194 µs for the Python module plus 26 µs for the `_operator` C accelerator), and it has no dependencies beyond that builtin.

PR #1160, opened 2026-09-01
fix: per-parameter limits in experimental mncontour and draw_mnmatrix off-by-one

henryiii:

`Minuit._experimental_mncontour` clamped both contour coordinates with the limits of the x parameter. If the two parameters had different limits, the y values could leave their allowed range. The fix clamps each coordinate with the limits of its own parameter. A new test checks that each parameter stays in its own range.

`Minuit.draw_mnmatrix` compared `y[k]` with `y[k - 1]` for `k = 0`, which wraps around to the last point and can add a wrong extreme. The loop now skips the first point.

The `mnprofile` docstring said the default of `size` is 100. The signature uses 30.

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -1698,7 +1698,7 @@ class Minuit:
         vname : int or str
             Parameter to scan over.
         size : int, optional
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
LGTM!

PR #1162, opened 2026-09-02
chore: use dependency groups instead of extras

henryiii:

Move the `test` and `doc` extras to PEP 735 `[dependency-groups]` and add a `dev` group that includes both. The extras were only used by `noxfile.py`, which now resolves the groups with `nox.project.dependency_groups` (requires nox>=2025.2.9). The contribute guide gets a short note on how to install a group without nox.

CI is unchanged: the docs and coverage workflows go through nox, and the test workflow installs packages explicitly.

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -49,8 +49,7 @@ If you prefer to manage the virtual environment yourself, the dependency groups 
 
 .. code-block:: bash
 
-    uv pip install -e. --group test
-    python -m pytest -n auto
+    uv run pytest -n auto
 
 Generate a coverage report:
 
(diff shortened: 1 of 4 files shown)


doc/contribute.rst
```diff
@@ -45,6 +45,13 @@ You have the source code now, next you want to build and test. We recommend to u
 
 This installs your version of ``iminuit`` locally and all the dependencies needed to run the tests, and then runs the tests.
 
+If you prefer to manage the virtual environment yourself, the dependency groups ``test``, ``doc``, and ``dev`` (both) in ``pyproject.toml`` list what you need. For example, with ``uv``:
+
+.. code-block:: bash
+
+    uv pip install -e. --group test
+    python -m pytest -n auto
```
henryiii:
```suggestion
    uv run pytest -n auto
```

ikrommyd (approved):
This is good but just a thought (for another PR), perhaps iminuit should provide optional dependencies for certain features like scipy for scipy minimizers or `PySide6` for example and ipywidgets for the interactive fitting mode.

PR #1164, opened 2026-09-02
fix: describe() accepts Annotated with non-float base type

henryiii:

`Annotated[int, Gt(0)]` or `Annotated[np.float64, Gt(0)]` crashed `describe()` and `Minuit` construction with `AssertionError`; the base type does not matter for limit extraction.

### DIFF
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -1377,8 +1377,7 @@ def _get_limit(
     if get_origin(annotation) is not Annotated:
         return None
 
-    # The base type (first argument) is irrelevant for limit extraction;
-    # accept any base type, not just float.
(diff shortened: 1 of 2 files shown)


src/iminuit/util.py
```diff
@@ -1377,8 +1377,9 @@ def _get_limit(
     if get_origin(annotation) is not Annotated:
         return None
 
-    tp, *constraints = get_args(annotation)
-    assert tp is float
+    # The base type (first argument) is irrelevant for limit extraction;
+    # accept any base type, not just float.
```
henryiii:
```suggestion
    # The base type (first argument) is irrelevant for limit extraction.
```

ikrommyd (approved):

PR #1165, opened 2026-09-02
fix: do not assert in FMin text repr when Hesse failed

henryiii:

Minuit2 can report `hesse_failed` together with `has_posdef_covar`, a combination the text repr did not expect. It asserted this could not happen, so `str(fmin)` raised `AssertionError` instead of printing a diagnostic message. Now this state degrades gracefully, showing "Covariance pos. def." alongside "Hesse FAILED".

### DIFF
--- a/src/iminuit/_repr_text.py
+++ b/src/iminuit/_repr_text.py
@@ -60,8 +60,10 @@ def fmin_fields(fm):
         covariance_msg1 = "Hesse FAILED"
         if fm.has_reached_call_limit:
             covariance_msg2 = "ABOVE call limit"
+        elif fm.has_posdef_covar:
+            # unexpected combination, but degrade gracefully instead of raising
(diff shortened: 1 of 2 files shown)


ikrommyd (commented):
I do understand that this is "fine" from a coding perspective but is it fine from a statistics/fitting perspective? Can we realistically reach this state some how? What is special about this state? Perhaps we should actually do something else than report back gracefully. I think this needs some stats research or @HDembinski's opinion because if this state means something is horribly wrong mathematically, gracefully reporting back is not correct.

henryiii:
This is for the string representation. I don't think `str(fmin)` should ever throw an error. If it reaches this state, something went wrong, but the string representation should not throw an error.

ikrommyd (approved):
Whoopsie, I did not notice it's for the repr only. Yeah sounds good then

PR #1166, opened 2026-09-02
fix: honor numpy boolean masks in parameter views

henryiii:

`m.values[mask]` (and similar indexing on parameter views) read a numpy
boolean array as integer indices, since `np.bool_` is not a subclass of
`bool`. An empty key raised `IndexError` instead of selecting nothing, and
`_ndim("ab")` looped forever because indexing a string yields another
string.


ikrommyd (approved):
LGTM!

PR #1167, opened 2026-09-02
fix: make hide_modules exception-safe and cheapen is_module_available

henryiii:

`hide_modules` removed its finder from `sys.meta_path` by identity, but did not guard against the finder already being gone, so an exception raised inside the `with` block (or code that rebuilt `sys.meta_path`) could raise `ValueError` on exit instead of restoring `sys.meta_path` and the hidden modules. `is_module_available` now checks for a module spec instead of fully importing the module, which is cheaper and avoids import side effects.

### DIFF
--- a/src/iminuit/_hide_modules.py
+++ b/src/iminuit/_hide_modules.py
@@ -46,5 +46,7 @@ def hide_modules(*modules, reload=None):
         yield
     finally:
         _forget(reload)
-        sys.meta_path.remove(finder)
+        # already gone if the block itself modified sys.meta_path
+        with contextlib.suppress(ValueError):
+            sys.meta_path.remove(finder)
         sys.modules.update(saved)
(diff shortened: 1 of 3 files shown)


ikrommyd (commented):
There's a merge conflict now so after you fix it just merge if you think it's right. It looks correct to me but I'm not very knowledgeable about the import system and you know a lot more than me here for sure.

henryiii:
`importlib.util.find_spec` is much better than trying to import, as it avoids the full import and just makes sure the module is importable. Due to Python's import system, it only works on a single level (for `a.b`, you can see if `a` is importable, or you can import a then see if `b` is importable, but you can't see if `a.b` is importable without importing `a`, that's not possible.) I haven't looked into the whole load/unload part of iMinuit (which I suspect might be better to remove; messing with this is tricky!), but the fix there looks reasonable.

PR #1169, opened 2026-09-02
ci: add Python 3.15 to tests and wheels

henryiii:

Adds a Python 3.15 test job on Ubuntu (numpy + scipy; matplotlib has no cp315 wheels yet) and `cp315`/`cp315t` to the wheel build matrix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-26-intel, macos-latest, ubuntu-24.04-arm, windows-11-arm]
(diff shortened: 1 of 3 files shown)


ikrommyd (commented):
Should probably completely drop 3.9 (your other PR) first right?

henryiii:
I was thinking we could do a final 3.9 release (which does include 3.15), then drop it, since all the development included it?

ikrommyd:
> I was thinking we could do a final 3.9 release (which does include 3.15), then drop it, since all the development included it?

Hmm sure I guess. I assume 3.15 you mean the rc and not actually wait for the release in a month.

henryiii:
The entire point of the RC is for packages to ship wheels. :)

ikrommyd (approved):
LGTM, let's just wait for workers to pick up all the jobs though.

PR #1170, opened 2026-09-02
fix: fall back to ndarray repr for non-square Matrix

henryiii:

Follow-up to #1139. `str()` and `_repr_html_()` on a `Matrix` assumed a square matrix and raised `IndexError` for 1D or non-square results of numpy indexing, such as `cov[np.array([0, 2])]` or `cov[0]`. They now fall back to the plain ndarray repr in that case.

### DIFF
--- a/src/iminuit/util.py
+++ b/src/iminuit/util.py
@@ -293,10 +293,13 @@ class Matrix(np.ndarray):
         else:
             self._var2pos = getattr(obj, "_var2pos", {})
 
+    def _is_square(self) -> bool:
+        return self.ndim == 2 and self.shape[0] == self.shape[1]
+
     def _names(self) -> Tuple[str, ...]:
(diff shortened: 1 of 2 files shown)


ikrommyd (commented):
Should we go with a fallback or actually implement similar pretty-printing/repr with the square case?

henryiii:
Non-square matrices don't have meaning, these are customized for covariance matrices. The idea was to just do show something default if this was sliced incorrectly, etc. So I think the fallback is correct.

ikrommyd:
> Non-square matrices don't have meaning, these are customized for covariance matrices. The idea was to just do show something default if this was sliced incorrectly, etc. So I think the fallback is correct.

ah yeah, make sense

ikrommyd (approved):

henryiii:
Slicing probably should not return the same subclass to make this harder to come across. But this is fine for now. :)

PR #1171, opened 2026-09-03
ci: build pyodide wheels

henryiii:

Adds `cp313`, `cp314`, and `cp315` pyodide (wasm32) wheel jobs to the release workflow, built with cibuildwheel and tested in node. The wheels are uploaded to PyPI with the others.

Two small source fixes were needed for wasm32: a sign-compare error in `flatten_hessian`, and `is_positive_definite` now handles a NaN result from cholesky, since pyodide's numpy returns NaN instead of raising.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,8 +75,13 @@ jobs:
 
   pyodide:
     needs: release_check
-    name: pyodide
+    name: ${{ matrix.py }} pyodide
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        py: [cp313, cp314, cp315]
+
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -111,9 +129,10 @@ jobs:
       attestations: write
 
     steps:
+      # pyodide wheels are not accepted by PyPI; they go on the GitHub release
       - uses: actions/download-artifact@v8
         with:
-          pattern: "*"
+          pattern: "{sdist,wheel-*}"
```
henryiii:
```suggestion
          pattern: "*"
```

ikrommyd:
Ah this may be cool for teaching. iminuit has been used and is really nice for giving lectures about statistics/fitting.

PR #1172, opened 2026-09-06
ci: build iOS wheels

henryiii:

Adds an `ios` job to the release workflow that cross-compiles wheels from both `macos-latest` and `macos-26-intel`, and includes them in the PyPI upload.

The `[tool.cibuildwheel.ios]` table uses the plain `build` frontend, declares `cmake` and `ninja` as `xbuild-tools`, and adds the beeware index, which supplies the iOS numpy wheels. Its environment table repeats the `*_ONLY_BINARY` settings because a platform table replaces the top-level one instead of merging it. The preinstalled Homebrew cmake cannot cross-compile to iOS, so the job reinstalls it.

Tests are skipped for iOS: CI has no simulator device.

Modelled on the equivalent job in boost-histogram.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,9 +114,6 @@ jobs:
 
       - uses: astral-sh/setup-uv@v10.0.1
 
-      # the preinstalled cmake cannot cross-compile to iOS
-      - run: brew uninstall cmake; brew install cmake
-
       - uses: pypa/cibuildwheel@v4.2
         env:
           CIBW_PLATFORM: ios
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,6 +98,34 @@ jobs:
           name: wheel-${{ matrix.py }}-pyodide
           path: ./wheelhouse/*.whl
 
+  ios:
+    needs: release_check
+    name: ${{ matrix.os }} ios
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, macos-26-intel]
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true
+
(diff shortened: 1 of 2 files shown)


ikrommyd (commented):
I've personally never ran python on my phone but sure, if those wheels are cheap to have I guess... The only thing I don't like about extra CI tasks is that there's yet another task to maintain and CIs are a thing that break or change over time.

Out of curiosity, is android possible too?

ikrommyd (approved):
I just checked and it's almost line-by-line identical to the boost-histogram job so it looks good to me. Passes in CI as well and had claude review it too.
You're definitely the ci person here out of the two of us @henryiii so just merge whenever you want.

henryiii:
> never ran python on my phone

I have, but this is so someone can make an app with Python using iMinuit in it.

> android

Yes, but there are general issues that are still being worked out, so I've been waiting until it's worked out. (Once https://github.com/scikit-hep/boost-histogram/pull/1045 is in, then doing it here too would make sense)

.github/workflows/release.yml
```diff
@@ -98,6 +98,34 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true
+
+      - uses: astral-sh/setup-uv@v10.0.1
+
+      # the preinstalled cmake cannot cross-compile to iOS
+      - run: brew uninstall cmake; brew install cmake
+
```
henryiii: This might not be required anymore. I'll try to check on boost-histogram and report here. (cmake 4+ is required, IIRC).

henryiii: https://github.com/scikit-hep/boost-histogram/pull/1189

henryiii: Yes, this is not needed anymore.

henryiii:
```suggestion
```

PR #1175, opened 2026-09-08
perf: do not sort in _unbinned_nll

henryiii:
This is a lot faster, but it is a significant change - sorting + pairwise is costly but drops the "more accurate" note. Needs further study, maybe it can be optional, etc. I'd rather expect a numba implementation possibly could use a more advanced summing algorithm (like `math.fsum`) instead of sorting?

Edit: keeping this in draft till after the next release, when it can be revisited.

`_unbinned_nll` sorted the log values before the sum, and with numba installed it dispatched to a jitted copy of the same code. Measurements against a `longdouble` reference show that the numba + sort path was the slowest option at every size and not more accurate than the plain `np.sum`, which uses pairwise summation.

| N | numpy + sort | numpy | numba + sort (old path) |
|---|---|---|---|
| 1e3 | 0.016 ms | 0.003 ms | 0.010 ms |
| 1e5 | 2.5 ms | 0.34 ms | 5.1 ms |
| 1e6 | 29 ms | 3.6 ms | 62 ms |

`UnbinnedNLL.__call__` with 200k points goes from 10.9 ms to 0.8 ms. The jitted variant is removed. There is now one implementation, so the numba/numpy parity test for this function is dropped.

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -164,8 +164,9 @@ def log_or_zero(x):
 
 
 def _unbinned_nll(x):
-    # sorting makes sum more accurate, protect against x = 0
-    return -np.sum(np.sort(np.log(x + _TINY_FLOAT)))
+    # adding tiny protects against x = 0; np.sum sums pairwise, which is accurate
+    # enough that pre-sorting is not worth its cost
+    return -np.sum(np.log(x + _TINY_FLOAT))
 
 
 def _z_squared(y, ye, ym):
@@ -445,14 +446,8 @@ try:
     def _ol_z_squared(y, ye, ym):
         return _z_squared  # pragma: no cover
 
-    _unbinned_nll_np = _unbinned_nll
-    _unbinned_nll_nb = jit(_unbinned_nll_np)
-
-    def _unbinned_nll(x):
-        if x.dtype in (np.float32, np.float64):
-            return _unbinned_nll_nb(x)
-        # fallback to numpy for float128
-        return _unbinned_nll_np(x)
+    # _unbinned_nll is not jitted: numba's naive sum is both slower and less accurate
+    # than the pairwise sum of np.sum.
 
     _multinomial_chi2_np = multinomial_chi2
     _multinomial_chi2_nb = jit(_multinomial_chi2_np)
--- a/tests/test_without_numba.py
+++ b/tests/test_without_numba.py
@@ -15,7 +15,6 @@ def npa(*args, **kwargs):
     (
         ("log_or_zero", (npa(1.2, 0),)),
         ("_z_squared", (npa(1.2, 0), npa(1.2, 1.0), npa(1.2, 1))),
-        ("_unbinned_nll", (npa(1.2, 0),)),
         ("multinomial_chi2", (npa(1, 0), npa(1.2, 0))),
         ("chi2", (npa(1.2, 0), npa(1.2, 1.0), npa(1.2, 0.1))),
         ("poisson_chi2", (npa(1, 0), npa(1.2, 0.1))),


henryiii:
Doing some exploration:

> For #1175, what about restoring the sort for the numpy path, and using a more advanced sumation algorithm (like math.fsum's alg) for the numba version?

Measured (`.nox/test`, normal pdf values, sum of already-computed logs; `log ms` is the cost of the `np.log` step that every variant pays):

| N   | method                        | sum ms | log ms | rel. error vs `math.fsum` |
| --- | ----------------------------- | ------ | ------ | ------------------------- |
| 1e5 | numpy sort + sum              | 2.14   | 0.18   | 0                         |
| 1e5 | numpy sum                     | 0.012  | 0.18   | 2e-16                     |
| 1e5 | numba naive                   | 0.058  | 0.18   | 5e-15                     |
| 1e5 | numba sort + naive (old path) | 4.71   | 0.18   | 6e-15                     |
| 1e5 | numba Neumaier                | 0.065  | 0.18   | 0                         |
| 1e5 | numba pairwise                | 0.017  | 0.18   | 0                         |
| 1e5 | numba fsum-style (Shewchuk)   | 0.81   | 0.18   | 0                         |
| 1e6 | numpy sort + sum              | 25.5   | 1.76   | 0                         |
| 1e6 | numpy sum                     | 0.13   | 1.76   | 0                         |
| 1e6 | numba Neumaier                | 0.65   | 1.76   | 0                         |
| 1e6 | numba fsum-style              | 8.6    | 1.76   | 0                         |

Full function (log + sum), numpy vs numba Neumaier vs numba pairwise: 0.195 / 0.189 / 0.193 ms at 1e5 and 1.89 / 1.90 / 1.94 ms at 1e6. The log dominates and numba gains nothing on it.

**Restoring the sort for numpy**: I would not. It costs 10 to 15x the entire rest of the function and buys at most one ulp, because `np.sum` is already pairwise. The per-element `log` has about one ulp of error itself, so the sum cannot be more accurate than that in any meaningful sense.

**A better summation for numba**: Neumaier or pairwise in a jitted loop is exactly rounded at the same cost as the naive loop, so it is strictly better than the old sort-then-naive path. An fsum-style partials loop is also exact but 13x slower than Neumaier, with no accuracy gain here. The catch is that a jitted variant has no speed advantage over plain numpy at all, so it only adds a second code path to maintain.

My recommendation stays with #1175 as is: one numpy implementation, no sort. If you want a guarantee of an exactly rounded sum rather than "within one ulp", the cheapest way is a jitted Neumaier loop for the numba path plus plain `np.sum` for numpy. I can put that on the branch if you prefer it.

---

>  Is there a manufactured case that pairwise sums really poorly compared to sorting + pairwise?

Yes, but it takes values that differ by more than 2^53 in magnitude, and it never gets worse than a handful of ulps. Errors below are in ulps of the exact sum (`math.fsum`), N = 1e6:

| case                               | pairwise | sort + pairwise | sort by \|x\| + pairwise | naive  |
| ---------------------------------- | -------- | --------------- | ------------------------ | ------ |
| A: one +1, rest +1e-16             | 7        | 0               | 0                        | 153102 |
| B: same, all negative              | 7        | 6               | 0                        | 153102 |
| C: 100 x +1, rest +1e-16           | 4        | 0               | 0                        | 7026   |
| D: 100 pairs of ±1e16, rest +1e-16 | 8e15     | 8e15            | 4e11                     | 8e15   |
| E: log N(0,1) at normal samples    | 0        | 1               | 1                        | 16     |
| F: 1e6 x ~-1e-9 and 1000 x -700    | 0        | 2               | 2                        | 1196   |
| G: 1e6 x log(tiny) and 10 x -1e-12 | 2        | 2               | 2                        | 25824  |

What that shows:

- **The manufactured win (A, C)** needs a few terms so large that the others fall below their ulp. Pairwise then loses the small terms that share a 128-element block with a big one, at most about 7 ulps. Sorting sums the small ones first and recovers them.
- **The sort in the old code was ascending by value.** For negative terms, which is what log-likelihood terms are, that puts the largest magnitudes first, so it does not fix case B (6 ulps left). Only a sort by magnitude helps, and that was never what the code did.
- **Cancellation (D)** is beyond any reordering; only compensated summation handles it. It cannot occur in an NLL sum because the terms have one sign.
- **Realistic NLL terms (E, F, G)** are within 0 to 2 ulps for pairwise, and sorting does not improve them. Log values live in roughly [-745, +700], so the magnitude ratio needed to trigger cases A or C requires terms near zero, and even case G stays at 2 ulps.

So the sort could only ever have mattered for a positive-valued sum with a 2^53 dynamic range, which the NLL never produces.

PR #1176, opened 2026-09-08
perf: store LeastSquares data in Fortran order for contiguous columns

henryiii:

`LeastSquares` passed strided column views of the `(N, ndim + 2)` data array to the model and the cost on every evaluation. The data array is now built with `np.stack(...).T`, which makes it Fortran-ordered, so the columns are contiguous views. Masked data is made Fortran-ordered again in `_update_cache`, because fancy indexing returns a C-ordered copy.

At N = 200k, the 1D `__call__` is 17% faster and the 2D `__call__` is 26% faster. Construction is about 3x faster as well, because `np.stack` writes one contiguous array instead of the strided copy that `np.column_stack` made.

Two new tests cover the `data` setter and 2D data with a mask. Both fail if the cache is stale.


ikrommyd (commented):
My AI agent informs me that 

> the cache silently breaks in-place edits: on develop c.x[0] = 10 then c.data[1,1] = 99 moves c(2) from 2.0 → 362.0 → 9386.0, and on the PR it stays 2.0 throughout. Only the setters invalidate, and the added test covers just c.data = ...; also worth noting it doubles the stored data.

henryiii:

Good catch, thanks. The cached copies are gone. The data array is now stored in Fortran order, so the columns are contiguous views of the data. In-place edits of `x`, `y`, `yerror` and `data` are visible again, and the data is not duplicated. Added a regression test for in-place edits.

The masked case was already a copy on develop (fancy indexing), so behavior there is unchanged.

henryiii:

Re-measured at the current head (3516d75) against `develop`. N = 200k, Python 3.13, numpy 2.5.2, numba 0.67, macOS arm64. Best of 10 repeats, two independent runs that agree to about 1%.

| benchmark | develop | this PR | change |
|---|---|---|---|
| 1D `__call__` | 0.333 ms | 0.283 ms | 15% faster |
| 2D `__call__` | 0.464 ms | 0.345 ms | 26% faster |
| 2D `__call__`, 50% mask | 0.229 ms | 0.170 ms | 26% faster |
| 1D construction | 0.178 ms | 0.328 ms | 1.8x slower |
| 2D construction | 0.309 ms | 0.578 ms | 1.9x slower |
| set `mask` (2D) | 1.02 ms | 1.03 ms | unchanged |

Two corrections to the description:

- 2D is 26% faster, not 34%. The 15% for 1D is correct.
- Construction is about 1.9x slower, because `np.asfortranarray` copies the array a second time. The cost is paid back after about two evaluations, so it does not matter for a fit, but it is not free.

The Fortran-order commit and the earlier cache commit measure the same on every call benchmark, so the rewrite kept the speed and dropped the extra memory.

henryiii:
Pushed a fix for the construction time, too.

henryiii:

Re-measured after the simplification to `np.stack(...).T` (b342133). Same setup as before: N = 200k, Python 3.13, numpy 2.5.2, numba 0.67, macOS arm64, best of 10 repeats, two independent runs agreeing to about 1%.

| benchmark | develop | this PR | change |
|---|---|---|---|
| 1D `__call__` | 0.342 ms | 0.283 ms | 17% faster |
| 2D `__call__` | 0.463 ms | 0.344 ms | 26% faster |
| 2D `__call__`, 50% mask | 0.231 ms | 0.171 ms | 26% faster |
| 1D construction | 0.183 ms | 0.062 ms | 2.9x faster |
| 2D construction | 0.307 ms | 0.089 ms | 3.4x faster |
| set `mask` (2D) | 1.03 ms | 1.03 ms | unchanged |

`np.stack` writes one C-contiguous `(ncol, N)` array, so the transpose is a Fortran-ordered view and no second copy happens. This is the same call speed as the earlier versions, and construction is now faster than `develop` instead of slower. The description still says 34% for 2D; the measurement is 26%.

`tests/test_cost.py` passes, and `data` keeps the dtype and contents the old `column_stack` produced.

ikrommyd (approved):

PR #1180, opened 2026-09-08
ci: set permissions, timeouts, and concurrency groups

henryiii:

- Add a top-level `permissions: contents: read` to all workflows. Only the job that creates the GitHub release gets `contents: write`. Pull request jobs no longer run with a write token.
- Add `timeout-minutes` to every job.
- Use `head_ref || run_id` in the concurrency groups of the coverage and docs workflows. `head_ref` is empty on push events, so all pushes shared one group and cancelled each other.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,12 +25,16 @@ on:
       - .github/workflows/coverage.yml
 
 concurrency:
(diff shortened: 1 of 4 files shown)


ikrommyd (approved):
Looks good to me. Just merge if you're ready. You know github CI much better than I do.

PR #1182, opened 2026-09-08
ci: use uv in the release workflow

henryiii:

The release check and sdist jobs used pip while the other workflows use uv. Switch them to `uv pip install --system` and drop `setuptools`, which scikit-build-core does not need.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,7 +138,7 @@ jobs:
           python-version: "3.12"
       - uses: astral-sh/setup-uv@v10.0.1
 
-      - run: pipx run build --sdist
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -135,11 +136,12 @@ jobs:
       - uses: actions/setup-python@v7
         with:
           python-version: "3.12"
+      - uses: astral-sh/setup-uv@v10.0.1
 
       - run: pipx run build --sdist
```
henryiii:
```suggestion
      - run: uv build --sdist
```

ikrommyd (approved):

PR #1183, opened 2026-09-08
chore: modernize CMakeLists.txt

henryiii:

- Raise the policy range to `3.15...4.0`. With CMP0148 set to NEW, pybind11 3 selects FindPython by default, so `PYBIND11_FINDPYTHON` is removed.
- Set `CMAKE_INTERPROCEDURAL_OPTIMIZATION` from `check_ipo_supported` before `pybind11_add_module`. pybind11 then does not add its own `pybind11::lto` target on top of the CMake IPO property, and the later `set_target_properties` call goes away.
- The compile definition on the MODULE target becomes `PRIVATE`.

Verified with a wheel build: FindPython is used, `-flto=thin` is applied once, and `tests/test_core.py` passes with the wheel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15...3.26)
+cmake_minimum_required(VERSION 3.15...4.0)
 
 project(iminuit LANGUAGES CXX)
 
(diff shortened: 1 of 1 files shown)


ikrommyd (approved):
You literally have a book about cmake :) You don't need my review.

PR #1189, opened 2026-09-08
feat: add optional dependency extras

### LINKED ISSUE #1168: Request: optional dependencies for certain features
This is good but just a thought (for another PR), perhaps iminuit should provide optional dependencies for certain features like scipy for scipy minimizers or `PySide6` for example and ipywidgets for the interactive fitting mode.

_Originally posted by @ikrommyd in https://github.com/scikit-hep/iminuit/pull/1162#pullrequestreview-5093601639_

henryiii:

Closes #1168.

Adds `[project.optional-dependencies]` extras: `scipy`, `plot` (matplotlib), `numba`, `ipywidgets` (+matplotlib), `qt` (PySide6 + matplotlib), `unicode` (unicodeitplus), and `all`. Documents them in the README and the install page.


pyproject.toml
```diff
@@ -31,6 +31,15 @@ classifiers = [
 ]
 dependencies = ["numpy >=1.21"]
 
+[project.optional-dependencies]
+scipy = ["scipy"]
+plot = ["matplotlib"]
+numba = ["numba"]
+ipywidgets = ["ipywidgets", "matplotlib"]
+qt = ["PySide6", "matplotlib"]
```
ikrommyd: this can potentially be grouped into one `interactive` optional dependnecy. Unless there's platform problems where they are not all available for all platforms.

pyproject.toml
```diff
@@ -31,6 +31,15 @@ classifiers = [
 ]
 dependencies = ["numpy >=1.21"]
 
+[project.optional-dependencies]
+scipy = ["scipy"]
+plot = ["matplotlib"]
+numba = ["numba"]
```
ikrommyd: perhaps `numba_stats` is related here too to get fast statistical distributions.

henryiii: We don't use it anywhere in iMinuit, though, so I'd leave it as a recommendation to users (though Claude agrees with you). Light and simple is good.

pyproject.toml
henryiii: :robot: _AI text below_ :robot:

It does differ: PySide6 has no wheels for PyPy, musllinux, iOS, or Pyodide, and it is a large download. Jupyter users would pull in Qt for nothing. My recommendation is to keep them separate, optionally adding `interactive = ["iminuit[ipywidgets,qt]"]` as a convenience alias.

henryiii: :robot: _AI text below_ :robot:

Added `interactive = ["iminuit[ipywidgets,qt]"]` as a convenience alias; `ipywidgets` and `qt` stay separate.

henryiii:
Maybe we should add minimums?

ikrommyd:
> Maybe we should add minimums?

yeah but I don't know them, it'd have to be done by testing

ikrommyd (approved):
I'd still be a fan of adding/promoting numba-stats too here as an extra. The tutorial notebooks use it.

henryiii:
That could be a followup. A "tutorials" or "examples" extra (we have one in some other packages) might make sense.

Issue #1192, opened 2026-09-08
Fable 5.1 xhigh review of iMinuit


henryiii:
This found a bunch of things, including #1190, which users ran into. Going ahead and posting it. The recent config/perf PRs were based on this.

At xhigh, there are more picky things, but there are some good things in here.

Short report followed by full report with numbering:

:robot: _AI text below_ :robot:

## Bugs worth fixing first

- **`log_or_zero` truncates int input** (`src/iminuit/cost.py:160`). Public `poisson_chi2`/`multinomial_chi2`/`template_chi2_*` return garbage for int counts, which is what `np.histogram` gives. Internal callers survive only because `_norm` casts first.
- **`mncontour(experimental=True)` returns complex points with numpy 2.5** (`src/iminuit/minuit.py:2680`). `np.linalg.eig` now always returns complex; the matrix is symmetric, so use `eigh`. The test suite already emits 21 `ComplexWarning`s that nobody sees.
- **Standalone `hesse()` fails when a parameter sits at 0 and fval != 0** (`src/functionminimum_extra.cpp:45`). The seed state drops the gradient step sizes, so Hesse uses a 3e-14 step. `Minuit(lambda x: (x-1)**2, x=0).hesse()` gives `hesse_failed=True`.
- **`import iminuit.qtwidget` fails with only PySide6 installed** (`src/iminuit/qtwidget.py:10`). Importing `backend_qt5agg` first forces Qt5 bindings. Tests pass only because pytest-qt imports PySide6 earlier.
- **`Minuit.g2` and `Minuit.hessian` raise `AttributeError`** (`src/iminuit/minuit.py:88`). The FCN binds `_g2`/`_hessian`; nothing tests these properties.
- **`FMin.time` goes negative after `hesse()`** (`src/iminuit/util.py:1047`). The timer subtracts the previous total instead of adding.
- **`describe()`** crashes on `Annotated[float, "doc string"]` (`util.py:1424`), includes keyword-only params so the fit call fails (`util.py:1302`), and only catches `NameError` from string annotations (`util.py:1397`).
- **`CostSum` loses `Annotated` limits depending on addition order** (`util.py:1153`). `nll + nc` keeps `sigma > 0`, `nc + nll` drops it.
- **Caller arrays mutated in place**: `NormalConstraint` squares the user's `error` array (`cost.py:2464`); `cost.data = ...` writes into the original data array (`cost.py:837`).
[...]

PR #1194, opened 2026-09-11
fix: accumulate FMin.time correctly across migrad/hesse/minos

henryiii:

`_Timer` in `src/iminuit/util.py` was meant to add elapsed time to the previous total, but `__exit__` computed `monotonic() - value` instead of `value + monotonic()`. For example, after `m.migrad()` followed by `m.hesse()`, `m.fmin.time` became negative instead of the sum of both runtimes. The fix makes `__enter__` subtract the start time and `__exit__` add the end time, so `value` accumulates correctly. Added a unit test for `_Timer` and an integration test that checks `fmin.time` grows after `migrad()` then `hesse()`.

Part of #1192


src/iminuit/util.py
```diff
@@ -1049,10 +1049,10 @@ def __init__(self, fmin):
         self.value = fmin.time if fmin else 0.0
 
     def __enter__(self):
-        self.value += monotonic()
+        self.value -= monotonic()
 
     def __exit__(self, *args):
-        self.value = monotonic() - self.value
+        self.value += monotonic()
```
ikrommyd: Is `perf_counter` perhaps a better clock for this? It's more precise and I was under the impression that `monotonic` is better for longer running tasks.

ikrommyd (approved):
It's correct, just one question about the clock.

src/iminuit/util.py
henryiii: On 3.13+, they are the same (at least on CPython). I think it might be better (the failure on Windows I think was before 3.13).

henryiii: I'm dropping the >=0 change, I think this fixes it before 3.13 on Windows.

ikrommyd (approved):
LGTM! Just fix the conflicts.

PR #1198, opened 2026-09-11
fix: first slider move after a fit in the ipywidget

henryiii:

`OnParameterChange` skipped one notification, because the code assumed that `slider.observe(...)` sends a notification when the observer is registered. traitlets does not do that. The first real slider move after a fit or a reset was therefore lost: `slider.value = 0.5` left `minuit.values[0]` unchanged and drew nothing, and only the second move worked.

`Parameter.reset` already removes the observer while it writes the new slider value, so programmatic updates cannot notify and the skip counter is not needed. Removed it and tightened the test: a single slider move after a fit or a reset must now update `minuit.values` and plot exactly once.

Part of #1192


henryiii:
Plotting seems to be a bit messy and this is probably the least of the problems. Let's push it off until we can properly look at polishing it.

ikrommyd:
I'm personally a fan of the interactive fitting. I was pushing for this in CMS for tag&probe measurements where the fits are hard to converge and depend strongly on the initial values. Seeing the 1D Z-peak fit interactively would help a lot. But yeah we can push it off for later cause I assume not many people care.

PR #1199, opened 2026-09-11
fix: use eigh for the symmetric covariance block in mncontour

henryiii:

`Minuit._experimental_mncontour` used `np.linalg.eig` on a symmetric 2x2 covariance block. numpy 2.5 changed `eig` to always return complex arrays, even for real symmetric input. This made `m.mncontour("x", "y", experimental=True)` return complex128 contour points, which triggers a `ComplexWarning` in `draw_mncontour`, or a `TypeError` in `MnUserParameterState.set_value` when warnings are errors. The fix uses `eigh`, which returns real eigenvalues for symmetric input, and reverses its ascending order to descending so the sampled contour points stay the same as before.

Part of #1192

### DIFF
--- a/src/iminuit/minuit.py
+++ b/src/iminuit/minuit.py
@@ -2677,12 +2677,18 @@ class Minuit:
 
         center = self.values[[ix, iy]]
         assert self.covariance is not None
-        t, u = np.linalg.eig(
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
`eigh` is faster too, not that it matters here

PR #1200, opened 2026-09-11
fix: keep fixed state in sync in the interactive widgets

henryiii:

The Fix and Fit toggles of the interactive widgets (`Minuit.interactive()`) did not keep `minuit.fixed` in sync. In the Qt widget, the sibling toggle was unchecked with `setChecked(False)`, which does not emit `clicked`, so a Fix click followed by a Fit click left `minuit.fixed[0] == True` while the Fix button showed unchecked. In both widgets, the fit handler restored the saved `fixed` state without `try/finally`, so an exception in the cost function left the override in place. Reset restored values and limits but not the fixed state or the Fix/Fit toggles, and a Fit toggle left on made Reset start a fit.

The Qt handlers now connect to `toggled(bool)`. The fit handler restores `fixed` in a `finally` block. Reset restores the fixed state recorded at construction and clears the Fix/Fit toggles with their signals blocked. Regression tests cover each case.

Part of #1192


ikrommyd (approved):
Just fix the conflicts and merge.

PR #1204, opened 2026-09-11
fix: avoid NaN in binned chi2 gradients when mu is zero

henryiii:

`_poisson_chi2_grad` and `_multinomial_chi2_grad` computed `n / mu` without a guard. If the model prediction underflows to exactly zero in a bin, this gives NaN or inf and the whole gradient becomes NaN with a RuntimeWarning. For example, `ExtendedBinnedNLL` with an exponential CDF at a small scale returns `[nan, nan]` from `grad`.

The value functions use `log_or_zero`, which is constant for `mu <= 0`, so the consistent derivative of the log term in such bins is zero. Both gradients now use a helper that evaluates `n / mu` only where `mu > 0`. The previously commented-out gradient checks in `test_ExtendedBinnedNLL_negative_weights` are restored, and a new test covers the multinomial path.

Part of #1192

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -285,9 +285,17 @@ def poisson_chi2(n: ArrayLike, mu: ArrayLike) -> float:
     return 2 * np.sum(n * (log_or_zero(n) - log_or_zero(mu)) + mu - n)
 
 
+def _n_over_mu_or_zero(n: NDArray, mu: NDArray) -> NDArray:
+    # log_or_zero is constant for mu <= 0, so the derivative vanishes in those bins
+    ma = mu > 0
+    r = np.zeros_like(mu * 1.0)
+    r[ma] = n[ma] / mu[ma]
+    return r
+
+
 def _poisson_chi2_grad(n: NDArray, mu: NDArray, gmu: NDArray) -> NDArray:
     assert gmu.ndim == 2
-    return 2 * np.sum((1.0 - n / mu) * gmu, axis=1)
+    return 2 * np.sum((1.0 - _n_over_mu_or_zero(n, mu)) * gmu, axis=1)
 
 
 def multinomial_chi2(n: ArrayLike, mu: ArrayLike) -> float:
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
I think this is good even though it's still draft.

henryiii:
The only downside is there's a runtime cost. Using `where=` helped, but it's not as fast as not doing anything. I'm exploring some fast path options.

`where=` is faster than masking at every size, roughly 1.4x for small arrays and 2.5x at 100k bins.

| bins   | plain `n/mu` (µs) | mask (µs) | `where=` (µs) |
| ------ | ----------------- | --------- | ------------- |
| 10     | 0.13              | 1.34      | 0.97          |
| 100    | 0.14              | 1.48      | 1.01          |
| 1000   | 0.27              | 3.00      | 1.64          |
| 100000 | 18.97             | 151.48    | 60.46         |

Both are several times slower than the unguarded division, mostly from the `zeros_like` allocation and the comparison. Given the gradient also does a broadcast multiply with `gmu` and a sum, this term is unlikely to dominate.

henryiii:
Two commits. One is a fast path that scans first. The second one catches the error and redoes it if there was a zero. That form doesn't handle negative values, but there shouldn't be any.

Full gradient with 3 parameters, µs per call, last column is the committed helper imported from `iminuit.cost`:

| bins   | plain | mask | `where=` | fast path | catch and retry |
| ------ | ----- | ---- | -------- | --------- | --------------- |
| 10     | 1.77  | 3.27 | 2.79     | 2.60      | 2.20            |
| 1000   | 3.20  | 6.60 | 4.92     | 4.34      | 3.76            |
| 100000 | 126   | 267  | 169      | 141       | 126             |

ikrommyd:
in my experience, catch and retry is usually the fastest in python (also when it comes to exception handling) if you are not hitting the exception too often. So if mu being zero is the rare edge case, I think catch and retry is best.

ikrommyd (approved):

PR #1206, opened 2026-09-11
fix: copy input arrays in cost constructors

henryiii:

Cost objects kept a reference to the arrays given by the user and then wrote into them. `err = np.array([1.0, 2.0]); NormalConstraint(["a", "b"], [0, 0], err)` left `err` as `[1., 4.]`, because the constructor squares the error array in place. In the same way, `x = np.array([1., 2., 3.]); c = UnbinnedNLL(x, pdf); c.data = [4, 5, 6]` changed `x`.

The constructors of `NormalConstraint` and `MaskedCost` now copy the arrays. `Template` already did this.

Part of #1192

### DIFF
--- a/src/iminuit/cost.py
+++ b/src/iminuit/cost.py
@@ -809,7 +809,8 @@ class MaskedCost(Cost):
         verbose: int,
     ):
         """For internal use."""
-        self._data = data
+        # copy, since the setters write into this array
+        self._data = data.copy()
         self._mask = None
         self._update_cache()
         Cost.__init__(self, parameters, verbose)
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):

ikrommyd:
I don't see any problem with this one @henryiii. It may only have excessive testing that's all. It would duplicate the double  for large arrays but if we write into them there's nothing we can do I guess. Deal with it however you want. I think this overlaps with https://github.com/scikit-hep/iminuit/pull/1176 in the least squares cost only in particular though.

henryiii:

Follow-up commit b102c01: `_norm` now has a `copy` flag, and the dtype conversion and the copy are one `np.array` call. Before, a non-float input was copied twice, once by `astype` and once by `.copy()`. The copy in `MaskedCost.__init__` is removed; `UnbinnedCost`, `BinnedCost`, and `NormalConstraint` pass `copy=True`.

`copy or None` is used because `copy=False` in NumPy 2 means "never copy" and raises when a dtype change is needed. Checked that `copy=None` behaves like `False` on NumPy 1.21.

src/iminuit/cost.py
```diff
@@ -809,6 +809,7 @@ def __init__(
         verbose: int,
     ):
         """For internal use."""
+        # data must be a fresh array, since the setters write into it
```
henryiii:
```suggestion
```

henryiii:
This should help reduce the number of copies in the case that the input is float. Now in either case there's one copy. Before the PR there was a copy (and no bug) if not float. In the first form of the PR non-float triggered two copies.

ikrommyd (approved):
Even better. LGTM!

PR #1207, opened 2026-09-11
fix: keep the seed gradient when hesse runs without a minimum

henryiii:

`Minuit(lambda x: (x - 1) ** 2, x=0).hesse()` set `hesse_failed` and gave no covariance. The same happened after `migrad()` when the user set a parameter back to 0 and called `hesse()`. The `init` function in `functionminimum_extra.cpp` built the `MinimumState` with an empty `FunctionGradient`, so `MnHesse` started from zero step sizes, which it clamps to about 3e-14 at a parameter value of 0. The fix passes `seed.Gradient()` into the `MinimumState`; `MinimumError(n)` stays unset so no covariance is claimed. With this change a user-supplied `hessian=` callable also takes the analytical path in this case (`nhessian > 0`).

Part of #1192

### DIFF
--- a/src/functionminimum_extra.cpp
+++ b/src/functionminimum_extra.cpp
@@ -42,7 +42,10 @@ FunctionMinimum init(const FCN& fcn, const MnUserParameterState& st,
   }
 
   MinimumParameters minp(val, err, seed.Fval());
-  std::vector<MinimumState> minstv(1, MinimumState(minp, seed.Edm(), fcn.nfcn_));
+  // keep the seed gradient; MnHesse uses its step sizes to start the numerical
+  // second derivatives, an empty gradient makes them collapse to near zero
+  std::vector<MinimumState> minstv(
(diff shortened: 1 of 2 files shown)


src/functionminimum_extra.cpp
```diff
@@ -42,7 +42,10 @@ FunctionMinimum init(const FCN& fcn, const MnUserParameterState& st,
   }
 
   MinimumParameters minp(val, err, seed.Fval());
-  std::vector<MinimumState> minstv(1, MinimumState(minp, seed.Edm(), fcn.nfcn_));
+  // keep the seed gradient; MnHesse uses its step sizes to start the numerical
+  // second derivatives, an empty gradient makes them collapse to near zero
```
henryiii:
```suggestion
```

ikrommyd (approved):

ikrommyd:
I leave the final merge to you in this one. Looks correct to me but not knowledgable enough about the bindings.

PR #1208, opened 2026-09-11
fix: make the MnPrint bindings memory-safe

henryiii:

Three problems in the `MnPrint` bindings. `MnPrint::Impl` indexed a four-element label array with the verbosity level, but `Verbosity` has a fifth level `Trace = 4`, which `MnLineSearch` and `InitialGradientCalculator` emit; `Minuit(lambda x: (x - 1) ** 2, x=0)` with `print_level = 4` read a garbage pointer and crashed. `PyMnPrint` held an `MnPrint` member, whose constructor and destructor push and pop a thread-local stack of raw prefix pointers, so out-of-order destruction in Python (`a = MnPrint("AAA", 3); b = MnPrint("BBB", 3); del a`) made `b.warn(...)` print the freed prefix of `a`. `show_prefix_stack`, `add_filter`, and `clear_filter` are static C++ functions, but were bound with `def`, so an instance call raised `TypeError`.

The fix adds the fifth label, stores only the prefix and level in `PyMnPrint` and makes a local `MnPrint` for each message so the push and pop stay nested, and binds the three static functions with `def_static`. The Python API does not change. Regression tests are in `tests/test_core.py`.

Part of #1192

### DIFF
--- a/src/print.cpp
+++ b/src/print.cpp
@@ -7,34 +7,49 @@ using namespace ROOT::Minuit2;
 using cstr = const char*;
 using namespace pybind11::literals;
 
-// MnPrint stores the prefix as a `const char*` and keeps it on a stack for its
-// whole lifetime. We therefore must keep the prefix string alive as long as the
-// MnPrint object exists. This holder owns the string; since members are
-// destroyed in reverse order of declaration, `print` is destroyed before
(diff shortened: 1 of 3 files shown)


ikrommyd (approved):
I was worried a bit that this may be a performance issue due to the fact that now construct the object per print/debug/etc statement but I measured and it's within noise.

PR #1212, opened 2026-09-14
fix: recompute total counts in BinnedNLL so in-place edits are visible

henryiii:
Astra noticed this while reviewing #1176.

`BinnedNLL` cached the total number of counts in `_update_cache` (#1149). In-place edits of `c.n[...]` on an unmasked histogram bypass the cache, so the prediction stayed at the old total and a perfectly matching cost became non-zero. The total is now computed from the live counts in `_pred` and `_grad`. Adds a regression test.


ikrommyd (commented):
Is there a performance hit here? This and other PRs I've seen about caching and in-place edits makes me thing there should probably be a cache that in-place operations invalidate? I don't think in-place operations are something users often do, that's why. I don't like overcomplicating code though.

henryiii:
Probably, that’s why this is still draft :). I’ll look into it a bit.

henryiii:

This reverts part of the caching from #1149. Measured cost of the uncached total per `BinnedNLL` call, with a trivial model `x**a` (worst case, the model is nearly free):

| bins | call | `np.sum` | `ndarray.sum()` |
|---|---|---|---|
| 10 | 3.1 µs | 14% | 6% |
| 100 | 3.7 µs | 12% | 5% |
| 1000 | 10.5 µs | 5% | 3% |

With a `scipy.stats.norm.cdf` model the share is 1 to 3% with `np.sum`, less with the method form. The second commit uses `ndarray.sum()`, which skips the `np.sum` dispatch overhead and halves the cost.

The cache cannot be kept only during fitting: NumPy has no hook for in-place writes, and `Minuit` has no start/end notification for the cost object. A read-only `n` would turn in-place edits into errors instead of making them visible.

ikrommyd (approved):

PR #1216, opened 2026-09-15
perf: declare lazy imports with flake8-lazy

henryiii:

Adds the [flake8-lazy](https://github.com/henryiii/flake8-lazy) pre-commit hook with `--apply=set`, scoped to `src/`, and runs it once. The hook writes `__lazy_modules__` declarations, which Python 3.15 honors as lazy imports (PEP 810, on by default) and older versions ignore.

After this, `import iminuit` no longer imports numpy: 57 ms -> 28 ms on 3.15 (hyperfine, macOS arm64). Bare interpreter startup is 13 ms.

The hook keeps the declarations correct as the code changes.


ikrommyd (approved):
Are we sure that none of these imports set up some global state that needs to be resolved at import time? If so, happy to just merge.

henryiii:

No import-time global state depends on any of the deferred modules, and the full suite passes on Python 3.15 with lazy imports active. But I found a real problem you should know about before merging: **shiboken6 (PySide6) breaks PEP 810 reification**, and iminuit ships a PySide6 widget.

**Global state audit.** The only module-level side effects in `src/iminuit` are the `try:` import blocks in `cost.py` (numba), `ipywidget.py`, and `qtwidget.py`. None were made lazy, and `iminuit._core` was left eager. `_deprecated.py` still evaluates `version("iminuit")` eagerly. Nothing registers filters, backends, or hooks at import time, so deferring the listed modules is semantically safe. `iminuit.util` and `iminuit.minuit` also remain reachable as package attributes on 3.15.

**Test results on 3.15.0rc2** (built in a scratch venv):

| Run                                      | Result                |
| ---------------------------------------- | --------------------- |
| default (pytest-qt autoloads PySide6)    | 22 failed             |
| `-p no:pytest-qt`, Qt test files ignored | 768 passed, 2 skipped |

**Cause.** shiboken6 replaces `builtins.__import__` with a C function `__feature_import__`. After that, lazy names are never reified: `iminuit.Minuit` stays a `lazy_import` object, so `from iminuit import Minuit` inside a function hands back the proxy (`'lazy_import' object is not callable`), and `warnings.warn(..., mutil.IMinuitWarning)` fails with `category must be a Warning subclass`. A pure-Python `__import__` override does not trigger it. Minimal repro with no iminuit at all:

```python
# pkg/__init__.py: __lazy_modules__ = {"pkg.sub"}; from pkg.sub import X
import shiboken6, pkg
type(pkg.X)   # <class 'lazy_import'>  (is <class 'type'> without shiboken6)
```

**Why it matters for iminuit.** Any user on 3.15 who imports PySide6 before iminuit, which is exactly the `interactive()` Qt workflow, gets a broken `iminuit`. This is an upstream bug (shiboken6 6.11.2 vs CPython 3.15rc2), but it will show up as iminuit bug reports, and CI on 3.15 will fail as soon as pytest-qt is installed.

Options, in the order I'd pick: keep the commit but report the bug to PySide6 and CPython and gate CI on 3.15 by disabling pytest-qt for now; or hold the commit until shiboken6 is fixed. The repro package is in the scratchpad if you want to file it.

henryiii:
Submitting this to Qt's Jira is annoyingly hard (my Qt email and Atlassian email don't match). We can wait on this a bit until it gets fixed. (It's also not our fault).

ikrommyd:
Okay then let's defer this one and you can just do the release.

